PR #231, opened 2014-12-27
Annoying notification about push access when viewing repositories. 

mattiasb:
I get a notification saying "must have push access to view repository collaborators (403)" whenever I view a repository. I guess it is true that I need this, but I don't need to be constantly informed (or at all). 

Thanks for an otherwise fantastic application! :)


ocdtrekkie:
I actually managed to backport this fix (and the other API fixes). #HoloForever

PR #377, opened 2016-12-28
fixed line breaks on release notes using newline characters

martino2k6:
It seems like the release notes on GitHub can use newline characters as opposed to standard HTML syntax. This commit makes sure that these get replaced by the appropriate HTML tag before being displayed in the TextView.


maniac103:
I assume you're referring to READMEs? Do you have an example of such a case?

martino2k6:
No, I mean the release notes. Try to open https://github.com/deltaDNA/android-sdk/releases/tag/4.2.6 in the app.

maniac103:
I think the root cause rather was wrong invocation of the markdown conversion. See above commit ;-)

martino2k6:
Pretty sure you know better than I do :)

I'll have a look over it when the fix will be out in a release, and re-open the issue if it won't be fixed. Thanks!

Issue #326, opened 2016-05-27
Remember last searches


maniac103:
From play store reviews: (Optionally?) remember last search keywords and auto complete them when user enters a new search

PR #395, opened 2017-02-04
Recent search suggestions

greexon:
Hi. I've implemented #326


maniac103:
Thanks for the PR, it's a long requested feature 👍  I wonder whether the suggestions shouldn't be split by search type, though, that is, if e.g. user search is selected, only show suggestions for previous user searches; rationale being the search input likely being completely different between those 3. It looks like this could be achieved by using 3 separate activity aliases in the manifest, having different searchable XMLs for each one, pointing to a custom suggestion content provider which distinguishes which one to use by the android:searchSuggestPath XML attribute.

maniac103:
We'll also need a 'clear search history' button.

greexon:
I've tried to do 3 different aliases but I have the following question: how should I change activity's context to setup correct `searchableInfo` in `searchView`?  I.e. when user selects search type in `spinner`, I need somehow setup new `searchableInfo` for `searchView` using `activity-alias`. Could you please advise?

Or maybe there is another more convenient way to implement this feature? Something like this: http://stackoverflow.com/questions/13733460/android-providing-recent-search-suggestions-without-searchable-activity

maniac103:
Reading this SO link, I think it's a much better approach for our use case. When designing the DB columns cleverly, one could probably also reuse it for providing suggestions for the issue search.
We'll need to make sure as much DB work as possible is offloaded to background threads though...that doesn't seem to be taken care of in the SO link.

greexon:
I've updated PR. Now custom `ContentProvider` is used instead of `SearchRecentSuggestionsProvider `

maniac103:
Looks much better. Threading is an issue, though. For the query part, I suggest using a Loader; for the insert part, I'd use an AsyncTask.
Why is there a need for multiple databases though? Why not simply extend DbHelper to take care of the new table?

greexon:
I've fixed threading issues and have extended `DbHelper`.

maniac103:
:+1: Will review and merge in the next days.

src/com/gh4a/activities/SearchActivity.java
```diff
@@ -155,6 +186,24 @@ public void onCreate(Bundle savedInstanceState) {
...
 
+        mSearch.setOnSuggestionListener(this);
+        mSearch.setSuggestionsAdapter(
+                new SimpleCursorAdapter(this, android.R.layout.simple_list_item_1, null,
+                        new String[]{SuggestionsProvider.Columns.SUGGESTION},
+                        new int[]{android.R.id.text1}, 0)
+        );
+
+        final SearchView.SearchAutoComplete complete = (SearchView.SearchAutoComplete) mSearch.findViewById(R.id.search_src_text);
+        complete.setThreshold(0);
```
maniac103: I'm not sure whether I'm a fan of this; autocompleting starting before the first character feels somewhat overzealous to me.  IMHO the default of 2 is better.

src/com/gh4a/activities/SearchActivity.java
```diff
@@ -155,6 +186,24 @@ public void onCreate(Bundle savedInstanceState) {
...
+        complete.setOnClickListener(new View.OnClickListener() {
+            @Override
+            public void onClick(View v) {
+                complete.showDropDown();
+            }
+        });
```
maniac103: Why is this needed?

src/com/gh4a/activities/SearchActivity.java
```diff
@@ -324,18 +375,46 @@ public boolean onQueryTextSubmit(String query) {
...
+    public boolean onSuggestionSelect(int position) {
         return false;
     }
 
     @Override
+    public boolean onSuggestionClick(int position) {
+        Cursor cursor = mSearch.getSuggestionsAdapter().getCursor();
+        if (cursor.moveToPosition(position)) {
+            mQuery = cursor.getString(1);
+            mSearch.setQuery(mQuery, false);
```
maniac103: I wonder whether we should submit right away here ... not sure though.

src/com/gh4a/activities/SearchActivity.java
```diff
@@ -375,4 +454,35 @@ public void onItemClick(Object item) {
...
+
+    private class SaveSearchSuggestionTask extends BackgroundTask<Void> {
+
+        public SaveSearchSuggestionTask() {
+            super(SearchActivity.this);
+        }
+
+        @Override
+        protected Void run() throws IOException {
+            saveSuggestion(mSearchType.getSelectedItemPosition(), mQuery);
```
maniac103: Type and suggestion should be passed into the task constructor instead of getting them from member variables to avoid threading issues.

src/com/gh4a/db/DbHelper.java
```diff
@@ -22,6 +23,12 @@ public void onCreate(SQLiteDatabase db) {
                 + "type integer not null, "
                 + "uri text not null, "
                 + "extra_data text);");
+        db.execSQL("create table " + SUGGESTIONS_TABLE + " ("
+                + "_id integer primary key autoincrement, "
+                + "type integer not null, "
+                + "suggestion text, "
+                + "date long, "
+                + "unique (type, suggestion) on conflict replace);");
```
maniac103: This also needs to be done in onUpgrade() (and the DB version also needs to be bumped). It's probably worth being put in a separate method, which is then called from both places.

maniac103:
Also (sorry I forgot to mention this earlier), can you please target this towards the 'printing' branch, which is the current development/feature branch?

maniac103:
I merged the commits manually and did the remaining fixups. Thanks for your work!

src/com/gh4a/activities/SearchActivity.java
greexon: This is required to show suggestions when user clicks on `SearchView`. But this code doesn't make sense if suggestions are shown after 2 symbols.

greexon:
Okey, thanks, but I wanted to do it by myself.

src/com/gh4a/activities/SearchActivity.java
```diff
@@ -128,6 +137,28 @@ protected boolean onError(Exception e) {
...
     };
 
+    private LoaderManager.LoaderCallbacks<Cursor> mSuggestionCallback = new LoaderManager.LoaderCallbacks<Cursor>() {
+
+        @Override
+        public Loader<Cursor> onCreateLoader(int id, Bundle args) {
+            return new CursorLoader(getApplicationContext(), SuggestionsProvider.Columns.CONTENT_URI,
+                    new String[]{SuggestionsProvider.Columns._ID, SuggestionsProvider.Columns.SUGGESTION},
+                    SuggestionsProvider.Columns.TYPE + " = ? and " + SuggestionsProvider.Columns.SUGGESTION + " like ?",
+                    new String[]{String.valueOf(mSearchType.getSelectedItemPosition()), mQuery + "%"},
```
greexon: I think it's required to use `Bundle args` here instead of using `mSearchType` and `mQuery` to avoid threading issues.

maniac103: I'll check this, but as far as I remember onCreateLoader is called from the main that.

PR #412, opened 2017-02-16
Issue/pull request action indicators

### LINKED ISSUE #398: Issue/pull request action comment indicators
Currently, there is no easy way to distinguish comments from issue/pull request actions such as close, merge, etc. What I propose is to add a small icon indicator near the user avatar which clearly shows that the specified comment has a special meaning. 

I did a small experiment to see what I could make and there is an example idea that I came up with:
image

If it's something that you would like to be added I can properly finish my work and create a pull request.

Tunous:
Links for easy testing:
https://github.com/Tunous/TestingRepo/issues/1
https://github.com/Tunous/TestingRepo/pull/2

Closes #398.


Tunous:
Note to self to fix this issue:
Event icon is not reset when rebinding view for different comment.

src/com/gh4a/adapter/IssueEventAdapter.java
```diff
@@ -156,6 +159,12 @@ protected Intent getIntent() {
             holder.tvDesc.setTag(null);
             holder.tvDesc.setText(formatEvent(event.event, event.getUser(),
                     holder.tvExtra.getTypefaceValue()));
+
+            Drawable issueEventIcon = getIssueEventIcon(event.event);
+            if (issueEventIcon != null) {
+                holder.ivIssueEvent.setVisibility(View.VISIBLE);
+                holder.ivIssueEvent.setImageDrawable(issueEventIcon);
+            }
```
maniac103: I know you noticed it already, just stating it explicitly:

```
} else {
    holder.ivIssueEvent.setVisibility(View.GONE);
}
```

is missing here.

src/com/gh4a/adapter/IssueEventAdapter.java
```diff
@@ -170,6 +179,59 @@ protected Intent getIntent() {
...
+                iconAttr = R.attr.issueEventRenamedIcon;
+                break;
+
+            default:
+                return null;
+        }
+
+        int iconResId = UiUtils.resolveDrawable(mContext, iconAttr);
+        return ContextCompat.getDrawable(mContext, iconResId);
+    }
```
maniac103: I wonder whether it would make sense to move the lookup into a static final HashMap<String, Integer>, so the binding code would become:

```
int eventIconAttr = EVENT_ICONS.get(event.getEventType());
if (eventIconAttr != 0) {
    holder.ivIssueEvent.setImageResource(UiUtils.resolveDrawable(mContext, eventIconAttr));
    holder.ivIssueEvent.setVisibility(View.VISIBLE);
} else {
    holder.ivIssueEvent.setVisibility(View.GONE);
}
```

maniac103:
Also, can you please target this towards the 'printing' branch? That one is the current development/feature branch; master will be used for bug fixes only. Yes, I suck at branch naming :-/

res/drawable/issue_event_closed.xml
```diff
@@ -0,0 +1,15 @@
+<vector
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportHeight="24.0"
+    android:viewportWidth="24.0">
+    <path
+        android:fillAlpha="1"
+        android:fillColor="#e65858"
```
Tunous: When I tried to use `@color` in vector drawables in past it didn't work. But, now I see a warning saying that the generated pngs might not have the correct colors. Perhaps they improved that since I last tried. I'll try that and test how it will look on API < 21.

I wonder if it would work when enabling compat vector drawables too. I use them successfully in my own app (there were some issues with proguard) but never tried using color references. If they indeed work then perhaps it's time to replace all the pngs :wink:

maniac103: I never got to look into compat vector drawables so far. If one can use those in menus etc., they sound like an even better choice, because one then could use theme attributes for the colors and get rid of the drawable duplication.

Tunous:
Rebased the changes to target printing branch.

res/drawable/issue_event_closed.xml
Tunous: Yeah this doesn't work correctly on API < 21: 
screenshot_20170217_122718

We should look into implementing support vector drawables in a separate pull request.

res/drawable/issue_event_unlocked.xml
```diff
@@ -0,0 +1,15 @@
...
+    android:viewportWidth="24.0">
+    <path
+        android:fillAlpha="1"
+        android:fillColor="#858585"
+        android:pathData="M12,-0C18.627,-0 24,5.373 24,12 24,18.627 18.627,24 12,24 5.373,24 0,18.627 0,12 0,5.373 5.373,-0 12,-0" />
+    <path
+        android:fillColor="#ffffff"
+        android:pathData="M19,16.2V19H16.2V16.9H14.1V14.8H12L10.418,13.218C10.033,13.337 9.627,13.4 9.2,13.4a4.2,4.2 0,0 1,-4.2 -4.2,4.2 4.2,0 0,1 4.2,-4.2 4.2,4.2 0,0 1,4.2 4.2c0,0.427 -0.063,0.833 -0.182,1.218L19,16.2M8.5,7.1a1.4,1.4 0,0 0,-1.4 1.4,1.4 1.4,0 0,0 1.4,1.4 1.4,1.4 0,0 0,1.4 -1.4,1.4 1.4,0 0,0 -1.4,-1.4z"
+        android:strokeWidth="0.69999999" />
+</vector>
```
maniac103: I wonder whether an open padlock wouldn't be more appropriate here?

maniac103:
Besides the one icon question, everything else looks great. Thanks so far!

res/drawable/issue_event_unlocked.xml
Tunous: Ah yes, I missed this one!

PR #416, opened 2017-02-17
Set default application theme

Tunous:
While the theme is set programmatically for each activity it is helpful to also define default theme for application in the AndroidManifest file. Doing so tells Android Studio to use this theme when looking at the layout preview. Without that Android Studio uses incorrect theme which requires to change it manually each time to be able to see correct preview and avoid errors.


maniac103:
Doing this will also change the theme used while starting up the application, before the first activity is loaded. Not to say that's a bad thing, though ;-)

PR #421, opened 2017-02-17
Do not animate drawer toggle icon

### LINKED ISSUE #402: Hamburger menu shouldn't animate when opening navigation drawer

Tunous:
Closes #402.


maniac103:
Shouldn't the drawer toggle rather be killed completely?

PR #424, opened 2017-02-17
Improve layout previews

Tunous:
Currently editing layout files is annoying because there is hardly anything displayed in the layout preview. This pull request fixes this by adding tools attributes to each of the layout files.

Before:
screenshot_20170217_184517

After:
screenshot_20170217_184542

It's **not a finished work** as I would like to know if I should continue with the rest of the layouts.

Additionally, I've added commit which sorted all layout attributes to appear in alphabetical order.

### DIFF
--- a/res/layout-land/color_picker_dialog.xml
+++ b/res/layout-land/color_picker_dialog.xml
@@ -5,13 +5,13 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:orientation="horizontal"
-    android:padding="@dimen/content_padding" >
+    android:padding="@dimen/content_padding">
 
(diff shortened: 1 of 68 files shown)


maniac103:
+1 for the preview attributes, -1 for the alphabetic sorting though: I've sorted all of them by keeping the layout attributes ('external' to the view) first, followed by android: ones and with app: ones last. If that general order is kept I'm fine with ordering the items alphabetically in those groups.

maniac103:
Looking at it again, it looks like your patch is following that order. Never mind then :)

Tunous:
It's done with default sort order of Android Studio: 
spectacle w14023

Tunous:
That was boring but at least now it'll be easier to work on layouts :smile:

PR #425, opened 2017-02-18
Comment actions

### LINKED ISSUE #423: Way to directly link to issue comment
On GitHub website, it's possible to get a direct link to issue comment by copying it from the timestamp. Here it could be implemented similarly by copying it to clipboard when clicking on the timestamp. 
<br />
<br />Alternatively, add 3 dots menu to each comment where the edit button is currently and list all comment actions here. This would allow for adding more actions such as open in browser, copy link, share, edit, etc.

Tunous:
Missing egit update to be able to use correct url.

There is a lot of code duplication between CommitNodeAdapter and IssueEvent adapter. Might want to refactor it.

Closes #423.


res/values/strings.xml
```diff
@@ -444,6 +444,8 @@
     <string name="share_commit_subject">Commit #%1$s at %2$s</string>
     <string name="share_user_subject">User %1$s (%2$s)</string>
     <string name="share_user_subject_loginonly">User %1$s</string>
+    <string name="share_comment_subject">Comment #%1$d at %2$s</string>
```
maniac103: 'Comment #%1$d in issue %2$d at %3$s'?

src/com/gh4a/adapter/IssueEventAdapter.java
```diff
@@ -62,10 +64,15 @@
         void editComment(Comment comment);
     }
 
+    private interface OnCommentMenuItemClick {
+        boolean onCommentMenuItemClick(Comment comment, MenuItem item);
+    }
```
maniac103: Shouldn't this interface be placed in the ViewHolder class?

src/com/gh4a/adapter/IssueEventAdapter.java
```diff
@@ -183,12 +222,14 @@ protected Intent getIntent() {
...
-            holder.ivEdit.setVisibility(View.VISIBLE);
-            holder.ivEdit.setTag(event.comment);
-            holder.ivEdit.setOnClickListener(this);
+            editMenuItem.setVisible(true);
+            editMenuItem.setEnabled(true);
         } else {
-            holder.ivEdit.setVisibility(View.GONE);
+            editMenuItem.setVisible(false);
+            editMenuItem.setEnabled(false);
         }
```
maniac103: Why disable hidden menu items?

src/com/gh4a/adapter/IssueEventAdapter.java
```diff
@@ -82,6 +89,32 @@
         EVENT_ICONS.put(IssueEvent.TYPE_RENAMED, R.attr.issueEventRenamedIcon);
     }
 
+    private final OnCommentMenuItemClick mCommentMenuItemClickCallback =
+            new OnCommentMenuItemClick() {
+                @Override
```
maniac103: Should indent by 4 here.

maniac103:
About the duplication, one can probably make a common base class, but I feel it'll become a generics hell (e.g. the menu item callback needs a common interface for Comment and CommitComment).

maniac103:
Looks good in general, only the icons need fixup: they should have the 'lighter' appearance (#343434/alpha 0.6; #ffffff/alpha 0.8) instead of black/white. Also I'd prefer to keep the icon naming in sync with the others (overflow.xml would be the grey one, overflow_dark.xml the white one).

Tunous:
Hmm, that would require adding additional an icon though. It's also used in the toolbar with full white color. How should I name it?

maniac103:
Rename the existing one to overflow_white and name the new ones overflow/overflow_dark?

Tunous:
Rebase to allow for merging.

Tunous:
To not forget about using correct url:

- [ ] Use correct url.

res/drawable/overflow.xml
```diff
@@ -0,0 +1,10 @@
+<!-- drawable/dots_vertical.xml -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
```
maniac103: I played with this and I think it would look better when using 18dp size here and using android:scaleType="center" in the ImageView displaying it. With the current size IMHO it looks a bit large.

Thoughts?

maniac103:
Pushed new bindings with the getHtmlUrl() method.

BTW: You know you can squash those little fixup commits into the main commit? That way one would get a cleaner commit history. I used squash on merge on your previous PRs, but this one contains commits that actually should be kept separate.

Tunous:
> BTW: You know you can squash those little fixup commits into the main commit? That way one would get a cleaner commit history. I used squash on merge on your previous PRs, but this one contains commits that actually should be kept separate.

Yeah, I am aware of this but I was keeping them separate to make reviewing easier. That way you could see exactly what changed since previous reviews.

1. I can stop doing that and keep only meaningful commits
2. You can post a comment after review for me to cleanup the branch

res/drawable/overflow.xml
Tunous: Seems good for me.

Tunous:
Anyway I cleaned this one.

maniac103:
Thanks. For the future I'd prefer option 1), if it's not too much effort. I tend to review the resulting code (as opposed to each PR iteration separately) anyway.

Tunous:
Ok, I'll be doing that from now :)

PR #430, opened 2017-02-19
Quoting

### LINKED ISSUE #252: Quoting
It would be nice if it was possible to select the text in issue and quote it without clipboard.

Tunous:
It works by adding an additional menu item to the actions menu when user selects text in comments. This allows for easy quoting of any part of the comment.

The quote insertion code appends the selected text at the end of the comment input field. It also makes sure that each quoted line is prefixed with quote char to appear correctly when rendered.

Additional code was added to make sure that there is at least one empty line above and under the quote text when adding it to comment input field.

Closes #252.

### DIFF
--- a/res/menu/comment_selection_menu.xml
+++ b/res/menu/comment_selection_menu.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/quote"
+        android:icon="@drawable/icon_quote"
+        android:title="@string/quote"
+        app:showAsAction="ifRoom" />
+</menu>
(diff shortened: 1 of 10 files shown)


Tunous:
Always forgetting about special cases ;(

TODO: Disable quoting if user can't comment.

src/com/gh4a/adapter/IssueEventAdapter.java
```diff
@@ -110,6 +112,10 @@ public void destroy() {
         mImageGetter.destroy();
     }
 
+    public void setLocked(boolean locked) {
+        mLocked = locked;
```
maniac103: notifyDataSetChanged() missing?

res/menu/comment_selection_menu.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/quote"
+        android:icon="@drawable/icon_quote"
```
maniac103: Is this icon actually ever shown anywhere?

maniac103:
Great stuff! I only wonder whether the icon is needed, everything else looks fine.

res/menu/comment_selection_menu.xml
Tunous: Yes on older API where the actions menu works like toolbar with icons. (I tested it on KitKat)

PR #454, opened 2017-02-22
Mentions

### LINKED ISSUE #447: Mention suggestions
When typing @ in comments or issue body display suggestions for users to mention. 

It might be possible to do with [AutoCompleteTextView](https://developer.android.com/reference/android/widget/AutoCompleteTextView.html) and adapter that overrides filtering behavior to only work when @ character is typed. 

As for the list of users to suggest then I'm not sure how GitHub does this but I think it's OK to just list repository collaborators and current issue/pr participants.

Tunous:
Implemented for the comment box in issue, pull request and commit comment views. It would be also nice to add this to issue creation screen but I can't think of the best way on how to get a list of participants. 

Another potential improvement is to also populate this list with possible repository assignees in addition to participants but I'm not sure if you want to make that additional request to the server.

Preview:
screenshot_20170222_201218

Closes #447.

### DIFF
--- a/src/com/gh4a/fragment/CommentBoxFragment.java
+++ b/src/com/gh4a/fragment/CommentBoxFragment.java
@@ -11,15 +11,19 @@ import android.text.Editable;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
-import android.widget.EditText;
+import android.widget.MultiAutoCompleteTextView;
 
 import com.gh4a.BaseActivity;
 import com.gh4a.ProgressDialogTask;
 import com.gh4a.R;
 import com.gh4a.utils.UiUtils;
+import com.gh4a.widget.DropDownUserAdapter;
 import com.gh4a.widget.SwipeRefreshLayout;
 
+import org.eclipse.egit.github.core.User;
+
 import java.io.IOException;
+import java.util.List;
 
 public class CommentBoxFragment extends Fragment implements
         View.OnClickListener, SwipeRefreshLayout.ChildScrollDelegate,
@@ -31,9 +35,10 @@ public class CommentBoxFragment extends Fragment implements
     }
 
     private View mSendButton;
-    private EditText mCommentEditor;
+    private MultiAutoCompleteTextView mCommentEditor;
     private Callback mCallback;
     private boolean mLocked;
+    private DropDownUserAdapter mAdapter;
 
     public void setLocked(boolean locked) {
         mLocked = locked;
@@ -63,6 +68,10 @@ public class CommentBoxFragment extends Fragment implements
         UiUtils.showImeForView(mCommentEditor);
     }
 
+    public void setMentionUsers(List<User> suggestions) {
+        mAdapter.replace(suggestions);
+    }
+
     @Override
     public void onAttach(Context context) {
         super.onAttach(context);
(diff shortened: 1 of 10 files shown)


maniac103:
Great stuff! Will look at it ASAP.

src/com/gh4a/utils/ApiHelpers.java
```diff
@@ -84,4 +87,25 @@ public static boolean loginEquals(User user, String login) {
...
+                    isUnique = false;
+                    break;
+                }
+            }
+            if (isUnique) {
+                uniqueUsers.add(user);
+            }
+        }
+
+        return uniqueUsers;
```
maniac103: I wonder whether we shouldn't simply give the User class an equals() method, we could use a Set with that addition.

src/com/gh4a/utils/UiUtils.java
```diff
@@ -406,4 +407,38 @@ public void onDestroyActionMode(ActionMode mode) {
 
         public abstract void onTextQuoted(CharSequence text);
     }
+
+    public static class WhitespaceTokenizer implements MultiAutoCompleteTextView.Tokenizer {
+        @Override
+        public int findTokenStart(CharSequence text, int cursor) {
+            int i = cursor;
```
maniac103: Couldn't you just directly use the cursor variable to decrement and return?

src/com/gh4a/fragment/CommentBoxFragment.java
```diff
@@ -63,6 +68,10 @@ public void addQuote(CharSequence text) {
         UiUtils.showImeForView(mCommentEditor);
     }
 
+    public void setMentionUsers(List<User> suggestions) {
+        mAdapter.replace(suggestions);
```
maniac103: Given this is called from a Loader callback, is mAdapter guaranteed to be != null here?

src/com/gh4a/widget/DropDownUserAdapter.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+        return mFilter;
+    }
+
+    private class ArrayFilter extends Filter {
+        @Override
+        protected FilterResults performFiltering(CharSequence constraint) {
+            FilterResults results = new FilterResults();
+
+            if (mOriginalUsers == null) {
```
maniac103: This - AFAICT - can't happen.

src/com/gh4a/widget/DropDownUserAdapter.java
```diff
@@ -0,0 +1,186 @@
...
+                synchronized (mLock) {
+                    mOriginalUsers = new ArrayList<>(mUsers);
+                }
+            }
+
+            if (constraint == null || constraint.length() == 0) {
```
maniac103: if (TextUtils.is empty(constraint)) {

maniac103:
Added some assorted nitpicks, other than that it looks just fine. I wonder whether the getUniqueUsers method - if we don't want to add the equals method - shouldn't simply be inlined in the base adapter's getUsers method (which then could be renamed appropriately).

src/com/gh4a/utils/ApiHelpers.java
Tunous: That's indeed a better option. I would recommend doing this for all the API models that have some sort of ID. It will make it easier to work with them in future.

Tunous:
Reminder: Test the popup in dark theme

maniac103:
Popup in dark theme is fine. Will do the nitpick fixes myself :)

PR #462, opened 2017-02-23
Notifications

### LINKED ISSUE #175: Implement GitHub Notifications
It should be really useful to add a panel with GitHub notifications, near the news feed.
The API seems simple to use: https://developer.github.com/v3/activity/notifications/

In a second round, it can help to solve #42.

Tunous:
Work in progress.

Still thinking about this but I tink I have a basic implementation idea. I want it to look similar to the event cards in "News Feed" but since each notification is a different item I'll have to do it a bit differently.

1. When the notification in a holder is null then we'll render a header with repository name and a button to mark all notifications from that repo as read. It should also have standing out background and a top shadow to replicate the card look.

2. When the notification is not null then we'll render a notification. This will have a card background and some more notification specific stuff on it. Not sure exactly how I'll make this one look like.

3. When the notification is set to be the last one it'll also render the bottom shadow which will, in turn, look like all the notifications and header were located on a single card.

You can get a basic idea of how the grouping will work if you open the app and check the log.

We'll see how it'll turn out once I get some more time to finish this. Let me know if you have any suggestions or questions.

Closes #175.

### DIFF
--- a/src/com/gh4a/adapter/NotificationAdapter.java
+++ b/src/com/gh4a/adapter/NotificationAdapter.java
@@ -11,11 +11,12 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
 import android.widget.PopupMenu;
+import android.widget.TextView;
 
 import com.gh4a.R;
 import com.gh4a.loader.NotificationHolder;
+import com.gh4a.utils.StringUtils;
 import com.gh4a.utils.UiUtils;
-import com.gh4a.widget.StyleableTextView;
 
 import org.eclipse.egit.github.core.Notification;
 import org.eclipse.egit.github.core.NotificationSubject;
@@ -96,7 +97,7 @@ public class NotificationAdapter extends
         }
 
         holder.ivIcon.setAlpha(alpha);
-
+        holder.tvTimestamp.setAlpha(alpha);
         holder.mPopupMenu.getMenu().findItem(R.id.mark_as_read).setVisible(!item.isRead());
 
         NotificationSubject subject = item.notification.getSubject();
@@ -113,6 +114,8 @@ public class NotificationAdapter extends
         }
 
         holder.tvTitle.setText(subject.getTitle());
+        holder.tvTimestamp.setText(StringUtils.formatRelativeTime(mContext,
+                item.notification.getUpdatedAt(), true));
 
         int bottomPadding = item.isLastRepositoryNotification() ? mBottomPadding : 0;
         holder.cvCard.setContentPadding(0, 0, 0, bottomPadding);
@@ -133,8 +136,9 @@ public class NotificationAdapter extends
             ivAction = (ImageView) view.findViewById(R.id.iv_action);
             ivAction.setOnClickListener(this);
             ivIcon = (ImageView) view.findViewById(R.id.iv_icon);
-            tvTitle = (StyleableTextView) view.findViewById(R.id.tv_title);
+            tvTitle = (TextView) view.findViewById(R.id.tv_title);
             cvCard = (CardView) view.findViewById(R.id.cv_card);
+            tvTimestamp = (TextView) view.findViewById(R.id.tv_timestamp);
 
             mPopupMenu = new PopupMenu(view.getContext(), ivAction);
             mPopupMenu.getMenuInflater().inflate(R.menu.notification_menu, mPopupMenu.getMenu());
@@ -143,8 +147,9 @@ public class NotificationAdapter extends
 
         private final ImageView ivIcon;
         private final ImageView ivAction;
-        private final StyleableTextView tvTitle;
+        private final TextView tvTitle;
         private final CardView cvCard;
+        private final TextView tvTimestamp;
         private final PopupMenu mPopupMenu;
         private final OnNotificationActionCallback mActionCallback;
 
--- a/src/com/gh4a/adapter/NotificationAdapter.java
+++ b/src/com/gh4a/adapter/NotificationAdapter.java
@@ -2,18 +2,22 @@ package com.gh4a.adapter;
 
 import android.content.Context;
 import android.content.res.Resources;
+import android.support.annotation.Nullable;
 import android.support.v7.widget.CardView;
 import android.support.v7.widget.RecyclerView;
 import android.view.LayoutInflater;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
+import android.widget.PopupMenu;
 
 import com.gh4a.R;
 import com.gh4a.loader.NotificationHolder;
 import com.gh4a.utils.UiUtils;
 import com.gh4a.widget.StyleableTextView;
 
+import org.eclipse.egit.github.core.Notification;
 import org.eclipse.egit.github.core.NotificationSubject;
 import org.eclipse.egit.github.core.Repository;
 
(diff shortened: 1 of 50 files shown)


maniac103:
Sounds good to me.

(Looks like loading the notifications takes quite a while for me - your log produces 676 lines ;-) )

Tunous:
Haha someone doesn't  check notifications ;)

There is one issue thought that I would also like to resolve but it doesn't seem to be possible now. It is marking notification for specific issue/pr as read whenever you open that issue. Similar to how it's done on GitHub. For that to be possible we need a `thredId` returned when getting the issue and I don't see it in the server response. Any way for you to contact GitHub people and ask them about that? Would also make it possible to subscribe to issues.

Tunous:
screenshot_1487888290

Suggestions for improvement? Right now it looks a bit too blank. I'll continue tomorrow.

maniac103:
It indeed looks a little lofty. I would maybe
- make the repo name use the HeaderLabel style
- make the notification rows a little flatter (maybe 32dp or something) [1]
- add icons for the subscription reason

[1] I know the touch target for the single notification clear icon becomes smaller than the guidelines say in that case, but my gut feeling is most people won't use it anyway, so I'm not sure whether we really should let this button dictate the overall layout.

Tunous:
screenshot_1487934385
I used subject type instead of notification reason to decide on the icon as this seems much more useful.

1. Is usage of fab to mark all notifications as read OK here? I can't decide between this or button in the toolbar menu.
2. Perhaps mark repository as read icon should be colored in green like the title too?
3. Are the other icons ok with these varying colors or it'll be better for them all to be grey?

Tunous:
I think I prefer the mark all as read in toolbar as fab overlaps with other stuff too much and adding padding to content looks bad...

maniac103:
> Is usage of fab to mark all notifications as read OK here? I can't decide between this or button in the toolbar menu.

The header aligned FAB definitely looks weird. An (auto hiding) FAB at the bottom would be ok, I think, but I think I prefer a toolbar menu item.

> Perhaps mark repository as read icon should be colored in green like the title too?

Not sure here. It would probably look nicer, but I'm not sure whether it would still be recognizable as a button? Also, we can't use icon tinting, can we?

> Are the other icons ok with these varying colors or it'll be better for them all to be grey?

Varying colors? I only see black ones :)
I think I'd prefer to have them grey instead of black to put more focus on the text, but I'll have to try this to tell for sure.

Tunous:
> Also, we can't use icon tinting, can we?

I'm not sure if it works for pngs but tinting vector drawables using DrawableCompat does work. 

> Varying colors? I only see black ones :)

I meant both icons on the left and right side.

maniac103:
> I meant both icons on the left and right side.

I don't see a problem there, as they have different purposes/semantics. I would suggest deciding on how much we want the left hand icons to stand out (see above).

Tunous:
TODO:

- [ ] Confirmation dialogs when marking repository or all notifications as read
- [ ] Save timestamp on notifications load for later use when marking as read to not mark all as read newer notifications
- [ ] Parse urls when clicking on notifications to redirect users to correct screens
- [ ] Add icons for all possible subject types (release, commit comment)

@maniac103 Might need help with parsing urls from `subject.url` in API as these might vary depending on the type. Definitely, there can be an issue and pull request types with links for them but I remember seeing on GitHub also different ones for releases and commit comments. Not sure if there is anything else...

Let me know how it looks so far. I think the UI/behavior needs some more work but I'm not sure what else could be done.

Tunous:
I figured out we can also display timestamps. This adds a bit more life to the layout.

screenshot_1488028947

Tunous:
Added some better link handling. It is going through existing BrowseFilter so all already supported links will be handled. I just had to make a small modification to it to normalize the links. What it means is that the BrowseFilter will now also open API links in application instead of browser. Is that ok?

maniac103:
Great work, LGTM :) A few minor comments:
- The 'mark all as read' icon IMHO shouldn't appear while the list is still loading
- The white space between icons and texts looks a little large; in particular it looks like the margins around the texts could simply be removed. Was there any particular reason for those?
- The notifications page should be added to the list of possible start pages in settings

> What it means is that the BrowseFilter will now also open API links in application instead of browser. Is that ok?
I don't understand that question. What do you mean by 'instead of browser'?

Tunous:
> Was there any particular reason for those?

I set this so the notifications text aligns with toolbar title. But you are right. It's a bit too much empty space. 

> What do you mean by 'instead of browser'?

OctoDroid doesn't currently handle api.github.com links but instead redirects to browser. After this change it'll normalize these links into regular ones and then open corresponding activity. I'm not sure about this change myself as someone could want to view api responses in browser and that could get annoying for them.

maniac103:
That one is totally not an issue in my POV. I can't imagine someone really clicks through the app and expects to get API responses opened in the browser.

Tunous:
Except it's not only clicking through the app but outside of it too.

Tunous:
My question is, should it be only handled when navigation in app or also outside it too.

maniac103:
It probably doesn't matter either way, as the total number of people looking at Github API responses on their Android device likely isn't far away from 0 ;)

I think I prefer to have this special URL conversion stuff only in the required places (here: notification fragment), though. Unless we have a specific need to change BrowseFilter, I'd suggest to keep its behaviour as-is.

Tunous:
Fixed. Might need some more testing to check if everything is working correctly but this pr should be ready now.

I think that there are missing some icons for few more notification types but I didn't see them in my notifications and didn't want to guess what they could be. *("Release"?)*

Tunous:
To check: How it looks on API < 21 as CardView shadows work differently here.

Tunous:
As I thought the shadows on pre-Lollipop don't work correctly:

screenshot_1488129598

I came up with this layout instead:

screenshot_20170226-182642

screenshot_20170226-183009

It doesn't look that bad - perhaps it could be used on all versions?

Tunous:
Pushed the commit with this change. Check it out and let me know which one you prefer.

maniac103:
In general it isn't too bad, but what irks me a bit is the missing shadow and/or spacing above the repo titles. If that could be fixed, that would be great.

Tunous:
I can see a shadow and spacing above repo titles. It's smaller than the bottom shadow but that's how CardView is designed.

maniac103:
Huh? I can't definitely see any spacing above the 'slapperwan/gh4a' header, only a divider line. Am I missing something?

Tunous:
That's the CardView top shadow. And I added additional 8dp margin to last notification to separate repositories, not sure if that was after I made the images though.

src/com/gh4a/adapter/NotificationAdapter.java
```diff
@@ -0,0 +1,212 @@
...
+                    (repository != null && item.repository.equals(repository)) ||
+                    (item.notification != null && item.notification.equals(notification))) {
+                item.setIsRead(true);
+            }
+
+            // When marking single notification as read also mark the repository if it contained
+            // only 1 notification
+            if (isMarkingSingleNotification) {
+                if (item.notification == null) {
+                    if (previousRepoItem != null && notificationsInSameRepoCount == 1) {
```
Tunous: Bug: This should be run only if the repository matches notification. Same for the final check below loop.

maniac103:
The margin is there, but it doesn't yield a top shadow :( As it stands, the look definitely is a bit weird, because the header's bottom shadows produce a 'stairs' like effect. If we can't make it work with CardViews in pre-L, I guess we'll have to drop the cards.

Tunous:
So I noticed that there is a weird issue with that top shadow. When scrolling I can notice that when these cards get higher on the screen then the shadow gets larger and smaller when they scroll down. The bottom shadow behaves in the opposite way. Perhaps the shadow is somehow cast from the center of the screen which makes it look different depending on the positon of the item. It feels like that's how elevation is done on Lollipop and higher and I don't really see a way how to change that...

Tunous:
An alternative that I can think of is to manually draw background and top/bottom shadows (gradient xml drawables) instead of using CardView. That would work on all API versions.

Tunous:
screenshot_20170301_182301

Is that good?

maniac103:
Yeah, that's good. The bottom edge of the cards should be fully white though, and the top shadow maybe a tad smaller than the bottom one.

Tunous:
It's 2dp vs 4dp for top and bottom shadow. Making top smaller will make it look like a divider.
Fixed the issue with bottom color in previous commit.

maniac103:
Well, original cardview top shadow *does* look like a divider, see your first screenshot. It should be fine if card background != activity background though.
So did you create fake shadow drawables or how did you solve it?

Tunous:
Yeah it's just a dark gradient as xml drawable for top shadow + the existing image for bottom shadow. Anyway, I changed it to 1dp. Looks ok for me.

maniac103:
Oh, I didn't notice you updated the PR. LGTM now, thanks a lot!

PR #469, opened 2017-02-26
Fix initial item highlight in navigation drawer

### LINKED ISSUE #466: Drawer items initial highlight issue
When users have set a page other than News Feed as their start screen, then the corresponding menu doesn't get highlighted when launching the application. Instead "News Feed" is always highlighted.

Tunous:
Closes #466.


maniac103:
Good catch. I think I'd prefer a callback style ('protected @IdRes int getInitialLeftDrawerSelection()') though, if only to keep symmetry with the other methods dealing with drawer creation. The default implementation could return 0.

Tunous:
Updated

PR #470, opened 2017-02-26
Fix toolbar collapsing when there are no items

### LINKED ISSUE #410: If there are no issues in issue list view header can't be scrolled to expand

Tunous:
Closes #410.


src/com/gh4a/fragment/LoadingListFragmentBase.java
```diff
@@ -84,6 +87,7 @@ protected void updateEmptyState() {
 
         mRecyclerView.setVisibility(empty ? View.GONE : View.VISIBLE);
         mEmptyView.setVisibility(empty ? View.VISIBLE : View.GONE);
```
maniac103: Can probably keep this view permanently visible now.

PR #473, opened 2017-02-27
Place issue title in HorizontalScrollView

### LINKED ISSUE #361: Handle long issue titles better
Currently, when viewing an issue, if its title is longer than two lines, it is just cut off. There is no indication that there is more text (e.g. an ellipsis) and no easy way to see the extra text, e.g. by clicking or scrolling, although turning to landscape works if it's not too long for that too.

See for example this issue (opened by me so my fault!)
screenshot_20161016-101846

Probably best to not make the header any larger so one fix would be to repeat the title in normal size text in the scrollable area below, e.g. above or below the username and date/time. Maybe only do this if necessary, i.e. when showing an ellipsis in the header title above.

Tunous:
Closes #361.

### DIFF
--- a/res/layout/issue_header.xml
+++ b/res/layout/issue_header.xml
@@ -5,7 +5,8 @@
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:orientation="horizontal"
-    tools:background="@color/issue_open_light">
+    tools:background="@color/issue_open_light"
+    tools:theme="@style/Theme.Header.Light">
 
     <com.gh4a.widget.VerticalTextView
         android:id="@+id/tv_state"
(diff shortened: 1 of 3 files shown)


Tunous:
peek 2017-02-27 19-37

Tunous:
peek 2017-02-27 19-45

maniac103:
This is an interesting idea. I wonder whether it's possible to theme the scroll bar, though, so that it's white and better fits into the header view.

maniac103:
Found what's off with the scrollbar: The header is inflated with the wrong theme. This diff should do the trick: https://gist.github.com/maniac103/2c63f37a743d250acd7da5e7ac3c2d76 ... can you incorporate it into the PR?

Tunous:
I'll look into that later today. Or you can also just push to this branch yourself. I enabled the option to allow edits from maintainers.

maniac103:
OK, I'll do that.

PR #474, opened 2017-02-27
Display confirmation dialog when closing/reopening issue

### LINKED ISSUE #129: Need a confirmation when closing an issue.
I can't tell you how many times I've accidentally closed an issue just by putting my phone in my pocket. It would be great if you could add a confirmation before actually closing it. It might eveb make sense to have it be a configurable setting for those that like the current behavior. Thanks!

Tunous:
Closes #129.


maniac103:
Are the dialog titles really needed? From my understanding of [1] they can be omitted here as they just repeat the message.

[1] https://material.io/guidelines/components/dialogs.html#dialogs-behavior

Tunous:
You are right. I based this one on the dialog used for removing bookmarks. I guess both should be changed.

PR #506, opened 2017-03-06
Lint fixes

Tommy-Geenexus:
This patchset deals with some of the Lint warnings, there are many more to resolve though.
I think it would make sense to adress all warnings in time, to avoid mountains of required overhauls in the future.

### DIFF
--- a/src/com/gh4a/utils/HtmlUtils.java
+++ b/src/com/gh4a/utils/HtmlUtils.java
@@ -305,9 +305,9 @@ public class HtmlUtils {
 
             // Fix flags and range for paragraph-type markup.
             obj = mSpannableStringBuilder.getSpans(0, mSpannableStringBuilder.length(), ParagraphStyle.class);
-            for (int i = 0; i < obj.length; i++) {
-                int start = mSpannableStringBuilder.getSpanStart(obj[i]);
-                int end = mSpannableStringBuilder.getSpanEnd(obj[i]);
+            for (Object span : obj) {
+                int start = mSpannableStringBuilder.getSpanStart(span);
+                int end = mSpannableStringBuilder.getSpanEnd(span);
 
                 // If the last line of the range is blank, back off by one.
                 if (end - 2 >= 0 && (end - start) >= 2) {
@@ -318,9 +318,9 @@ public class HtmlUtils {
                 }
 
                 if (end == start) {
-                    mSpannableStringBuilder.removeSpan(obj[i]);
+                    mSpannableStringBuilder.removeSpan(span);
                 } else {
-                    mSpannableStringBuilder.setSpan(obj[i], start, end, Spannable.SPAN_PARAGRAPH);
+                    mSpannableStringBuilder.setSpan(span, start, end, Spannable.SPAN_PARAGRAPH);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


maniac103:
The current development branch is the 'printing' branch, please rebase your changes onto that ... master currently carries bug fixes only. I believe I fixed a couple of those there already.

I've avoided following a number of those lint suggestions on purpose, though, because IMHO they do more harm than good to the understanding of the code. Among those are
- the method return value inversion ones (I strongly prefer a meaningful semantics of the return value over the potential always-inverted usage of the return type - a good example of that is the downloadNeedsWarning method, which after your changes returns false if a download needs a warning)
- the expression simplification ones (I prefer simpler logic statements if possible to make it easier to follow the logic)
- the package private access scope (it seems pointless/arbitrary as Octodroid doesn't have many packages and doesn't provide a public API, so I'm using only public = accessible for all, protected = only to be used by descendant classes, private = private, which provides a clear semantics. Limiting access between those three (public -> protected) is fine though.

Tommy-Geenexus:
Sure i'll rebase.
It would make sense though to supress specific Lint warnings then to make it clear for others not to work on them.

maniac103:
If there's a way to globally disable them, sure, go ahead. I don't think it's a good idea to add a suppression e.g. to each protected method though.

Tunous:
While we are at it, it would be also a good idea to add the android studio code style rules file (`.idea/codeStyleSettings.xml`) to git repository.

Tommy-Geenexus:
@maniac103 
Pr updated. I will adress other warnings in another pr, as i need to perform full rebase.
A question besides: Which branch should i build off to get the most recent updates if i want to build a stable version of the app, master or printing?

maniac103:
@Tommy-Geenexus Please use printing. The only reason I haven't merged it back yet is that a last bug fix release for Gingerbread and ICS users is supposed to be released from it.

src/com/gh4a/utils/HtmlUtils.java
```diff
@@ -305,9 +305,9 @@ public Spanned convert() {
 
             // Fix flags and range for paragraph-type markup.
             obj = mSpannableStringBuilder.getSpans(0, mSpannableStringBuilder.length(), ParagraphStyle.class);
-            for (int i = 0; i < obj.length; i++) {
-                int start = mSpannableStringBuilder.getSpanStart(obj[i]);
-                int end = mSpannableStringBuilder.getSpanEnd(obj[i]);
+            for (Object anObj : obj) {
```
maniac103: I suggest naming this 'span' (because that's what it is).

Tommy-Geenexus:
Thanks for clarification, i've updated the commit.
Besides: Would you be interested in something like [https://github.com/Applications-Development/gh4a/commit/b81b50908f39139ab993254e6c8a3e85661cd908](url)?
The idea is to enable auto searching for e.g. users ,and could eventually be extended to other types (though this commit may not be best fit, just to showcase the idea).

maniac103:
As for the question, I'm not sure. In general I'm fine with the idea of on-the-fly search, but we'd need to limit search result count for it (is this possible with the API? didn't check yet...), as otherwise it would just be a waste of bandwidth. I don't really like the implementation, though: query text change should only load a limited subset of results (as mentioned above) (+ show a 'load more' view to fetch the full list), while submitting should load the full list.

Tommy-Geenexus:
You're right.
I didnt really dig deeper while writing this, or cared about bandwidth (as i have unlimited),
but for upstreaming this it must be overhauled.
If there would be API support i think it would be a good idea to add on the fly search to certain activities,
as its much more comfortable for the user.

PR #509, opened 2017-03-07
Add scrollbars to screens with scrollable content

Tunous:


maniac103:
Why only here and not e.g. in the repo or user overview? Also, there surely must be a reason this is not on by default?

Tunous:
> Why only here and not e.g. in the repo or user overview?

I didn't notice it in these screens. Will fix them too. 

> Also, there surely must be a reason this is not on by default?

Seeing that the scrollbars code is implemented inside of View class I would guess it's disabled by default because ScrollView doesn't change it by itself and having it enabled by default for View is definitely not a good default.

Tunous:
Updated for more screens

Issue #409, opened 2017-02-16
Incorrect line numbers when adding line comment in diff view


Tunous:
Often diffs contain folds which hide many lines but this behavior is not respected by the line display for line comments. It is always counted based on a number of visible lines.

maniac103:
Got an example?

maniac103:
Oh, right. Issue is here: https://github.com/slapperwan/gh4a/blob/master/src/com/gh4a/activities/DiffViewerActivity.java#L237 ... probably need to parse the line numbers out of the diff to calculate an offset :-/

maniac103:
Actually parsing the numbers shouldn't be too hard. Every @@ line contains line numbers for old and new file, which then need to be incremented by one for old file (--- lines) or new file (for +++ lines) or both (for context lines). Depending on whether the comment is on a --- or +++ line, one can pass the appropriate number into the URL. Just need to decide which of the two numbers to use for context lines.

Tunous:
I think it would be a nice idea to change the format of that message to display both numbers. Something like "Commented on line -274, +276" similar to the format used on the fold.

PR #515, opened 2017-03-11
[WIP] Direct link to file/line in commits and pull requests

### LINKED ISSUE #413: Direct link to line in commit doesn't work correctly
Instead of opening code view for the selected file with the specified line highlighted it is only opening commit screen and doesn't go further than that. 

Example: https://github.com/slapperwan/gh4a/commit/c021f4585244ba4a4cfdb90203aec1034818a47c#diff-1e105dc4c4e8c142976247ec34a5f610R275

Tunous:
Added it both for the links to files in commits and pull requests.

@maniac103 Need your help with line highlighting. I don't see it being implemented for file diffs and am not really sure how to do it correctly.

Closes #413.

### DIFF
--- a/src/com/gh4a/activities/DiffViewerActivity.java
+++ b/src/com/gh4a/activities/DiffViewerActivity.java
@@ -231,10 +231,10 @@ public abstract class DiffViewerActivity extends WebViewerActivity implements
             }
 
             int pos = mHighlightIsRight ? rightDiffPosition : leftDiffPosition;
-            if (pos == mHighlightStartLine) {
+            if (pos != -1 && pos == mHighlightStartLine) {
                 highlightStartLine = i;
             }
-            if (pos == mHighlightEndLine) {
+            if (pos != -1 && pos == mHighlightEndLine) {
                 highlightEndLine = i;
             }
 
--- a/src/com/gh4a/activities/DiffViewerActivity.java
+++ b/src/com/gh4a/activities/DiffViewerActivity.java
@@ -64,15 +64,19 @@ import java.util.List;
 public abstract class DiffViewerActivity extends WebViewerActivity implements
         View.OnTouchListener {
     protected static Intent fillInIntent(Intent baseIntent, String repoOwner, String repoName,
-            String commitSha, String path, String diff,
-            List<CommitComment> comments, int initialLine) {
+            String commitSha, String path, String diff, List<CommitComment> comments,
+            int initialLine, int highlightStartLine, int highlightEndLine,
+            boolean highlightisRight) {
         return baseIntent.putExtra("owner", repoOwner)
                 .putExtra("repo", repoName)
                 .putExtra("sha", commitSha)
                 .putExtra("path", path)
                 .putExtra("diff", diff)
                 .putExtra("comments", comments != null ? new ArrayList<>(comments) : null)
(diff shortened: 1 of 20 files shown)


Tunous:
A bug I found that might be ok to fix in this pull request:

Navigating up from pull request file diff goes to pull requests list instead of pull request.

maniac103:
@Tunous I hope you don't mind me updating this branch. Please let me know what you think of this solution.

src/com/gh4a/activities/DiffViewerActivity.java
```diff
@@ -192,15 +206,36 @@ protected String generateHtml(String cssTheme, boolean addTitleHeader) {
...
             } else if (line.startsWith("+")) {
+                ++rightDiffPosition;
                 cssClass = "add";
             } else if (line.startsWith("-")) {
+                ++leftDiffPosition;
                 cssClass = "remove";
+            } else {
+                ++leftDiffPosition;
+                ++rightDiffPosition;
+            }
```
maniac103: The logic I added here should also allow fixing #409.

Tunous:
It's ok but that 'duplicated fetch logic' was done on purpose. With it users are redirected directly to the diff activity which has these advantages:

- They don't see commit activity while the commit is still loading
- Instead they see proper loading dialog
- Pressing back redirects them back to where they were instead of commit activity (they can go to commit using up arrow)

IMO this behavior is much better than opening 2 activities. It results in better back stack behavior and less screen jumps.

maniac103:
I think I do the same thing when going to a specific release though ... adapting that means adding more duplicated logic to BrowseFilter, which is something I really want to avoid to get into (because there's not only the code duplication, but also a high chance of subtle behavior differences AKA subtle bugs). I also tend to think of the URL fragments as 'initial navigation hints', not as something that completely should change the activity a using is seeing ... so I guess we have different views on this topic. :-/

maniac103:
After some more thinking I think you're right though. I'm just wondering whether showing only the diff viewer provides enough context, as its title bar info is rather sparse.
Feel free to revert back to your version of the code, otherwise I'll do it tomorrow. I'd like to keep the pre-parsing of the fragment before starting a background task, though.

Tunous:
https://github.com/Tunous/TestingRepo/releases/tag/0.0.1

Tunous:
I think that these load tasks together with the progress dialog and a new base class (for the common dialog showing/error handling code) should be moved outside of `BrowseFilter`. Perhaps to their own package each as a separate file.

Tunous:
> I'm just wondering whether showing only the diff viewer provides enough context, as its title bar info is rather sparse.

I don't think that's a problem. Users can just press on the up arrow and see the matching commit with more information.

maniac103:
Why should those tasks be moved out? AFAICT all of them are specific to loading stuff from URLs - at least the RefPathDisambiguationTask surely is.

Tunous:
No real reason for that. Just a simple suggestion since they seem to be more for direct launching of different screens than parsing URLs. (Except for the RefPathDisambiguationTask)

maniac103:
👍 from my side now. @Tunous are you also fine with it?

Tunous:
Yep :+1:

maniac103:
Shipit then, thanks for your patience with my back-and-forth thoughts :)

PR #516, opened 2017-03-11
Switch to correct pull request tab if it was linked directly

### LINKED ISSUE #456: Open in browser for pull request file diff links incorrectly
1. View pull request
2. Go to files tab 
3. View diff for a file 
4. Click on the open in browser action 

Currently doing so opens latest pull request commit diff in browser when instead it should open the files diff for the whole pull request. Better yet if it scrolled to that file but that needs [the SHA we don't know](https://github.com/slapperwan/gh4a/issues/413). (Or we do in this case?) 

Example link: https://github.com/slapperwan/gh4a/pull/454/files#diff-ce9de4f775f8605b5ba5a0a0dd200ac3

Tunous:


maniac103:
@Tunous I adapted the logic to what's already present in RepositoryActivity. OK with this?

src/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -249,7 +253,7 @@ protected void onInitExtras(Bundle extras) {
         mRepoName = extras.getString("repo");
         mPullRequestNumber = extras.getInt("number");
         mInitialCommentId = extras.getLong("initial_comment", -1);
-        mTargetScreen = extras.getString("target_screen");
+        mInitialPage = extras.getInt("initial_page", -1);
         extras.remove("initial_comment");
         extras.remove("target_screen");
```
Tunous: "initial_page"

maniac103: Oops, right.

Tunous:
OK now :+1:

PR #520, opened 2017-03-13
Fix new issue not appearing in list after creation

### LINKED ISSUE #458: Adding new issue doesn't show it in the list immediately
Requires a reload for it to show up.

Tunous:
Closes #458.


Tunous:
Still breaks sometimes so this fix won't work :(

PR #521, opened 2017-03-13
Correctly link pull request files

### LINKED ISSUE #456: Open in browser for pull request file diff links incorrectly

Tunous:
Closes #456.

### DIFF
--- a/src/com/gh4a/activities/CommitDiffViewerActivity.java
+++ b/src/com/gh4a/activities/CommitDiffViewerActivity.java
@@ -22,6 +22,7 @@ import android.support.v4.content.Loader;
 import com.gh4a.Gh4Application;
 import com.gh4a.loader.CommitCommentListLoader;
 import com.gh4a.loader.LoaderResult;
+import com.gh4a.utils.ApiHelpers;
 
 import org.eclipse.egit.github.core.CommitComment;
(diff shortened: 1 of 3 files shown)


maniac103:
Hmm, this feels a bit inconsistent. The commit URL doesn't have the path, while the PR one does. I'm not sure what version is actually more correct, but I feel they should be consistent.

maniac103:
I also wonder whether createUrl() shouldn't be abstract in DiffViewerActivity.

Tunous:
> The commit URL doesn't have the path, while the PR one does.

Oh yes. I think that the path should be added to commit URL too. That way it will behave like on GitHub web.

> I also wonder whether createUrl() shouldn't be abstract in DiffViewerActivity.

Sure.

Tunous:
Updated.

PR #523, opened 2017-03-13
Fix login not working if activity gets killed

Tunous:


src/com/gh4a/activities/Github4AndroidActivity.java
```diff
@@ -103,9 +111,11 @@ protected void onNewIntent(Intent intent) {
                     .appendQueryParameter(PARAM_CODE, data.getQueryParameter(PARAM_CODE))
                     .build();
             new FetchTokenTask(uri).schedule();
-        } else {
-            super.onNewIntent(intent);
+            return true;
         }
+
+        super.onNewIntent(intent);
```
maniac103: This line should be removed, right? Otherwise the super method is called twice AFAICT.

Tunous: Ops forgot to delete ;)

Tunous: Fixed

PR #526, opened 2017-03-15
Fix off by one error for organization selection

Tunous:
The `position == 0` is used for viewing user activity which makes it necessary to distract 1 from the position when attempting to retrieve correct organization


maniac103:
*facepalm* Thanks.

Issue #530, opened 2017-03-17
Markdown file rendering issues


Tunous:
[Example file](https://github.com/Tunous/TestingRepo/blob/master/markdown-test.md), compare how it looks in app and in browser.

Issues I found:

- No line breaks between paragraphs
- Some characters aren't rendered correctly, e.g.: ’
- Unnecessary line breaks (See [emphasis](https://github.com/Tunous/TestingRepo/blob/master/markdown-test.md#emphasis) section)
- In dark theme tables have wrong backgrounds

maniac103:
The above fixes item 2.

maniac103:
For item 3, my stance is that Github's rendering is wrong, as those line breaks are present in the Markdown file. Also, can you please explain what exactly you mean in item 1?

Tunous:
For item 3, looking at [the specification](https://github.github.com/gfm/) [released recently](https://githubengineering.com/a-formal-spec-for-github-markdown/) their rendering isn't wrong. There are 2 versions of line breaks and this one isn't supposed to start a new line. (The difference is the presence of 2 or more spaces before a line break)
 https://github.github.com/gfm/#soft-line-breaks
https://github.github.com/gfm/#hard-line-breaks

Bluscream:
Tables also look like shit. Is there no other 3rd party MD library you could use?

maniac103:
Feel free to point us towards one that works better than showdownjs. Also 'look like shit' isn't really meaningful feedback.

Bluscream:
# Syntax guide
Here’s an overview of Markdown syntax that you can use anywhere on GitHub.com or in your own text files.

## Headers

# This is an &lt;h1> tag
## This is an &lt;h2> tag
###### This is an &lt;h6> tag

## Emphasis

*This text will be italic*
_This will also be italic_

**This text will be bold**
__This will also be bold__

_You **can** combine them_

## Lists

### Unordered

* Item 1
* Item 2
  * Item 2a
  * Item 2b

### Ordered

1. Item 1
1. Item 2
1. Item 3
   1. Item 3a
   1. Item 3b

## Images

![GitHub Logo](/images/logo.png)
Format: ![Alt Text](url)

## Links

http://github.com - automatic!
[GitHub](http://github.com)

## Blockquotes

As Kanye West said:

> We're living the future so
> the present is our past.

## Inline code

I think you should use an
`<addr>` element here instead.

# GitHub Flavored Markdown

[...]

PR #536, opened 2017-03-20
Comment bottom sheet

Tunous:
Work in progress but ready for testing and first review.


res/drawable/collapse_dark.xml
```diff
@@ -0,0 +1,10 @@
+<!-- drawable/minus_circle_outline.xml -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="24dp"
+        android:height="24dp"
+        android:viewportHeight="24"
+        android:viewportWidth="24">
+    <path
+        android:fillColor="#99343434"
```
maniac103: #ccffffff

Tunous:
I think that's a good idea. Markdown buttons could be moved to the bottom then.

maniac103:
It might also be worth a thought to color the tab strip in black-on-white to avoid changes in background color of toggle and send button.

Tunous:
I personally prefer the look of colored tabs. The changes in background color shouldn't be a problem as I don't think users will often toggle between the modes more than once. Actually, maybe it's better to just not let them leave the advanced mode once they enter it once?

What's worth considering is replacing the advanced editor with a ViewStub and inflating it only when user enters advanced mode.

maniac103:
This is how it looks with b&w tabs, FWIW:
screenshot_20170320-155446
I kinda prefer it, as there's less visual distraction.

👍 for the ViewStub, not sure about the 'don't allow leaving advanced mode' bit. I guess I'm fine with it if needed, as it doesn't impair usability anyway.

BTW, something seems to be off in simple mode:
screenshot_20170320-155502

Tunous:
Hmm I thought it would look worse. I guess that's fine then.

Tunous:
Perhaps markdown buttons could get colored background:

screenshot_20170320-193636

maniac103:
Yeah, that looks OK.

Tunous:
screenshot_20170320-193844

That was wrong color :)

Tunous:
This happens sometimes when selecting text (or just clicking, I'm not sure). Can't find where that could be happening...

```
android.view.InflateException: Binary XML file line #43: Binary XML file line #43: Error inflating class TextView
Caused by: android.view.InflateException: Binary XML file line #43: Error inflating class TextView
Caused by: java.lang.UnsupportedOperationException: Failed to resolve attribute at index 5: TypedValue{t=0x2/d=0x1010435 a=1}
at android.content.res.TypedArray.getColorStateList(TypedArray.java:528)
at android.widget.TextView.<init>(TextView.java:1076)
at android.widget.TextView.<init>(TextView.java:704)
at android.support.v7.widget.AppCompatTextView.<init>(AppCompatTextView.java:60)
at android.support.v7.widget.AppCompatTextView.<init>(AppCompatTextView.java:56)
at android.support.v7.app.AppCompatViewInflater.createView(AppCompatViewInflater.java:103)
at android.support.v7.app.AppCompatDelegateImplV7.createView(AppCompatDelegateImplV7.java:1008)
at android.support.v7.app.AppCompatDelegateImplV7.onCreateView(AppCompatDelegateImplV7.java:1067)
at android.support.v4.view.LayoutInflaterCompatHC$FactoryWrapperHC.onCreateView(LayoutInflaterCompatHC.java:44)
at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:769)
at android.view.LayoutInflater.createViewFromTag(LayoutInflater.java:727)
at android.view.LayoutInflater.rInflate(LayoutInflater.java:858)
at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:821)
at android.view.LayoutInflater.rInflate(LayoutInflater.java:861)
at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:821)
at android.view.LayoutInflater.rInflate(LayoutInflater.java:861)
at android.view.LayoutInflater.rInflateChildren(LayoutInflater.java:821)
at android.view.LayoutInflater.inflate(LayoutInflater.java:518)
at android.view.LayoutInflater.inflate(LayoutInflater.java:426)
at android.view.LayoutInflater.inflate(LayoutInflater.java:377)
at android.widget.Editor$SuggestionsPopupWindow.initContentView(Editor.java:3343)
at android.widget.Editor$PinnedPopupWindow.<init>(Editor.java:3025)
at android.widget.Editor$SuggestionsPopupWindow.<init>(Editor.java:3309)
at android.widget.Editor.replace(Editor.java:356)
at android.widget.Editor$3.run(Editor.java:2129)
at android.os.Handler.handleCallback(Handler.java:751)
at android.os.Handler.dispatchMessage(Handler.java:95)
at android.os.Looper.loop(Looper.java:154)
at android.app.ActivityThread.main(ActivityThread.java:6209)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:865)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:755)
```

Tunous:
Added markdown buttons but there is a problem with keyboard covering them... Not sure what's the best way to fix that.

maniac103:
Probably need to adjust their bottom padding by adjusting to remaining screen space.

Tunous:
Didn't manage to get anything working with your solutions but I tried moving the markdown buttons out of scrolling container and it worked. The only difference is that they are now also visible when switching to preview tab.

I don't have any better ideas right now so I'm leaving it to you.

maniac103:
So how about this? It seems to work for me.

maniac103:
One open issue is the simple editor: when entering the 3rd line, it hides behind the keyboard instead of scrolling up the bottom sheet (which one can do manually). It's also seemingly not possible to scroll in the basic editor.
In the advanced editor it would be nice if scrolling in the editor would be possible without scrolling the app bar, but I'm not sure whether that's possible?

Tunous:
The first issue might be solvable by wrapping the basic editor in a nested scroll view. 

The second issue is something I tired to fix too but didn't find a working solution yet...

Tunous:
Also noticed that the editor doesn't collapse after sending a comment. We should fix that.

maniac103:
The second issue could probably be fixed by a callback mechanism that ultimately calls setToolbarScrollable(false) while the editor has focus.

Tunous:
The comment preview is no longer scrollable.

maniac103:
Hmm, for the preview I tried both putting the WebView into a NestedScrollView and https://github.com/takahirom/webview-in-coordinatorlayout/blob/master/app/src/main/java/com/github/takahirom/webview_in_coodinator_layout/NestedWebView.java#L96 ... both don't work though, as 'something' is intercepting touch events and the WebView gets ACTION_CANCEL events during the movement :(

maniac103:
Actually, scrolling behaviour is pretty screwy in when scrolling the advanced editor as well: when scrolling it the app bar scrolls as well and when reaching the editor top/bottom the sheet scrolls as well, while either shouldn't happen.

maniac103:
Guess what, this actually works :)

Tunous:
Text ****

Tunous:
Above comment shows a bug with strikethrough. It appears with that strikethrough line in preview tab when it shouldn't.

Tunous:
More markdown improvement task for myself:

- [ ] Make code tag insertion intelligently use inline or block tags
- [ ] Improve positioning of cursor after inserting element
- [ ] Implement url tag insertion similar to how GitHub does it
- [ ] Surround nearest text when using surround tags without selection (see GitHub)

Tunous:
https://github.com/slapperwan/gh4a/pull/536#issuecomment-287891497

Could this be something with context using incorrect theme?

maniac103:
> Could this be something with context using incorrect theme?

Yeah, in particular missing theme attributes. I've never seen this myself so far though, how do you repro this?

Tunous:
I have no idea how to fully reproduce it. I was just working with markdown buttons and that happens randomly when clicking somewhere inside of input field.

Tunous:
We should also use this new editor for line comments.

Tunous:
Some more issues:

- Expanded state is not remembered on device rotation
- The content in advanced editor is ignored and only what was in basic editor is restored after rotation
- When rotating from landscape mode to portrait there is a blank space instead of editor

maniac103:
I guess we'll need a better layout for landscape as well: https://drive.google.com/file/d/0B0hmLAh3l4oHeE1Jcno0R01lMzQ/view?usp=drivesdk

Tunous:
An alternative to solve the issue with toolbar auto hiding could be to force it to collapsed size and disable expanding when the advanced editor is open.

Tunous:
The only change for the landscape layout that I can think of right now is to move the title input field to the scrolling container. And change the title of tab to "Edit" instead of "Body".

maniac103:
Or allow scrolling out the toolbar with the tabs and move the markdown buttons to the right...
:+1: for the force-collapse toolbar with advanced editor idea.

Tunous:
> Or allow scrolling out the toolbar with the tabs and move the markdown buttons to the right

Not sure about scrolling out the toolbar but the idea with moving markdown buttons is nice.

Tunous:
We are almost there! It slowly starts to work nicely :)

maniac103:
Added some fixes for landscape layout. There's two issues with it, though; I'd appreciate if you could have a look at them:
- [ ] When leaving the advanced editor in landscape mode, the editor sheet is scrolled out completely; there's only a grey area where the basic editor should be. Touching the list view makes it appear again
- [ ] Shadow of markdown buttons in landscape mode is nonexistent. I didn't find a solution for the shadow as it would need to span both the left and top edge now.

Tunous:
> When leaving the advanced editor in landscape mode, the editor sheet is scrolled out completely; there's only a grey area where the basic editor should be. Touching the list view makes it appear again

Can't find a way to fix this. It happens only in pull request view and is caused by toolbar not being scrollable.

Tunous:
https://github.com/slapperwan/gh4a/pull/536#issuecomment-287891497

This is caused by spell check suggestions popup. If I type non-existing world like "Sfsfsfs " and click on it after red underline appears the app crashes. No ideas how to fix that though...

maniac103:
Works for me?

Tunous:
Android 7.1.1, emulator - nexus 5X
Android 7.0, OnePlus 3T

Happens on both of them

maniac103:
Hmm, tried both with simple and advanced editor and both light and dark theme - all of those worked. Oppo R7 Plus, LineageOS 14.1, thus Android 7.1.1

Maybe I am missing something, can you list exact reproduction steps?

maniac103:
In any case, the issue is some faulty/unexpected android:textColor attribute.

Tunous:
peek 2017-03-24 21-21

Maybe it's a compilation error. I wanted to record gif and all worked correctly so I recompiled and it broke again as you can see in the above gif...

Tunous:
Forgot to mention but I tried the NestedScrollView approach for basic editor earlier and it didn't work due to 2 issues. 

- I failed with setting max height so it was taking whole screen. This might be perhaps solvable by changing something with height or adding custom nested scroll view with max height. 
- The scroll views in advanced mode stopped working. No idea why this happened...

You might want to give it a try or we have to look for another solution.

Tunous:
Another issue I found: Settings tab in issue edit activity sometimes doesn't appear.

I believe this happened because of my current internet connection issues but there should be at least some information saying that loading went wrong.

maniac103:
> Another issue I found: Settings tab in issue edit activity sometimes doesn't appear.
> I believe this happened because of my current internet connection issues but there should be at least some > information saying that loading went wrong.
Exceptions during loader loads will yield the 'loading data from github failed' error message. The settings page needs push access to the respective repo though - might you simply not have had that?

maniac103:
> Forgot to mention but I tried the NestedScrollView approach for basic editor earlier and it didn't work due to 2 > issues.

Why do we even need a scroller there? AFAICT it should be sufficient to move the sheet up when the basic editor expands, wouldn't it?

Tunous:
> The settings page needs push access to the respective repo though

No, I had that access. It all loaded correctly once my connection issues were fixed. But there was no error message visible anywhere, just the whole settings page was gone.

> AFAICT it should be sufficient to move the sheet up when the basic editor expands, wouldn't it?

It also needs to be possible to scroll inside of basic editor when someone types text longer than 3 lines.

maniac103:
> The scroll views in advanced mode stopped working. No idea why this happened...

Probably the basic scrollview consumed the scroll? Might be fixable by disabling it or setting it to GONE.

Tunous:
Issue: I think that the title input field in issue edit screen has unnecessary bottom margin and wastes some space.

maniac103:
Right. Due to the tabs we don't need to add the padding that compensates for the FAB anymore, at least not in portrait mode.

Tunous:
Made the basic editor scrollable but... it has a similar issue with toolbar scrolling as the advanced editor.

Tunous:
So, what do we have left to do:

- [ ] I still didn't find a fix for the suggestion popup crash as I can't find what exactly could be causing it. This definitely needs to be fixed before merging.
- [ ] The CSS styling for previews should be modified to stay as close to the result as possible. We also have to fix few issues like these mentioned in https://github.com/slapperwan/gh4a/pull/536#issuecomment-288467111 and #530.
- [ ] New bug I found: When going to commit and switching to comments tab, the bottom sheet appears only after starting to scroll vertically.
- [ ] More testing ;)

Anything else that I missed?

Also, I think I might move the code from `MarkdownUtils` to a separate library. I have a few potential improvements for it and it could be used in other applications too as it's pretty generic. Switching to it could be always done at a later time even after this pull request is merged. Are you ok with that?

maniac103:
I'm not sure how I could help with the crash issue, ad I have yet to see it myself :-/ I'm really wondering what's going on there. About the preview rendering, I'd keep rendering differences aside for now, that is, separate from this PR. We can debug the showdown setup separately.
About MarkdownUtils I don't have an overly strong opinion either way. If it's generic enough to be put in a separate library, just go ahead. That reminds me that I wanted to do the Github binding library inclusion in a sane way (that doesn't involve copying a library - e.g. using a submodule or something) since ages :-/

Tunous:
> About the preview rendering, I'd keep rendering differences aside for now, that is, separate from this PR. We can debug the showdown setup separately.

That sounds ok to me.

> About MarkdownUtils I don't have an overly strong opinion either way. If it's generic enough to be put in a separate library, just go ahead.

Ok cool. Right now it's just an idea, so if/when I make it then I'll just make a separate pull request.

Tunous:
I've finally found a cause and solution to the crash :D

Time for a story ;p

___

To find out where the issue started I decided to run `git bisect` and find the first broken commit. It turned out to be 14347045f9b34be52848f3e54d7ebf896355f7aa. Based on my experiments of commenting out the code to see when the issue disappears/reappears I found the exact cause which is present in this file: https://github.com/slapperwan/gh4a/commit/14347045f9b34be52848f3e54d7ebf896355f7aa#diff-cbc41660757ea9338d9058aeaa8590f5. It doesn't look like anything bad happened here. It's just a WebView added to the layout file. It turns out that **this is the cause!** Having WebView present near editor crashes the app when suggestions popup appears.

To confirm this further I switched back to the latest code and commented out the WebView and everything dealing with it. This indeed fixed the issue. But that's not what we want to do. WebView is necessary to implement comment preview! So, having localized the cause and seeing that it is most likely not a fault of the app I began browser searching.

After some time I found [this issue](https://code.google.com/p/android/issues/detail?id=221504) on the bug tracker. Here they mention that the issue is fixed in v25.0.0 of support library. I upgraded the app to use the latest one and the crashes stopped!

There aren't any reasons to not update it right? Otherwise, we'll have to find a workaround or not use the WebView approach.

___

That's it. I just had to write about that ;)

maniac103:
Unfortunately, there are reasons: https://github.com/slapperwan/gh4a/commit/1a037ccf6df280bebd503e9c8c3bc62e7e2325f8 :( I believe we should be able to work around that issue, just need to find a good 'anchor' to place that additional requestLayout call.

Tunous:
> Are you sure this doesn't have any unintended consequences? I'm not sure whether there's additional stuff we have to support by targeting N.

Looking at the [listed behavior changes](https://developer.android.com/about/versions/nougat/android-7.0-changes.html) the only thing that could be problematic is this:

> Many platform APIs have now started checking for large payloads being sent across Binder transactions, and the system now rethrows TransactionTooLargeExceptions as RuntimeExceptions, instead of silently logging or suppressing them. One common example is storing too much data in Activity.onSaveInstanceState(), which causes ActivityThread.StopInfo to throw a RuntimeException when your app targets Android 7.0.

This could result in crashes in places where many commit comments are passed through Bundle like here: https://github.com/slapperwan/gh4a/blob/326edebc584ab5aa167926910406c2cd01e385f7/src/com/gh4a/fragment/CommitNoteFragment.java#L60.

Other than that I don't see anything needing changes.

Tunous:
> Unfortunately, there are reasons: 1a037cc :( I believe we should be able to work around that issue though, just need to find a good 'anchor' to place that additional requestLayout call.

Is this still happening with the latest code? I tried to reproduce it but for me, everything works correctly.

Tunous:
Another few bugs in CommmitNoteFragment:

1. Toolbar gets stuck in locked mode with laggy jumps when scrolling:

- Go to CommitNoteFragment
- Enter advanced mode
- Leave advanced mode
- Try to scroll comments list

___

2. Bottom sheet disappears when trying to scroll on it when in landscape mode (in CommitNoteFragment).

Tunous:
Scrolling down in basic editor when reached top tries to expand toolbar which results in weird behavior.

Tunous:
Got the FAB issue but it looks slightly different

 fab

maniac103:
Rebased branch onto master.

Tunous:
Just for your information: I'm using the app on my device with this branch for testing purposes for a few weeks already and I stopped seeing any new bugs. This means that it should be ready to merge once we resolve the final issues.

Unfortunately, so far I didn't manage to find any working fixes by myself :(

The last bugs I see are these unchecked from here:
- https://github.com/slapperwan/gh4a/pull/536#issuecomment-290815919
- https://github.com/slapperwan/gh4a/pull/536#issuecomment-291005759

maniac103:
Maybe I'm misunderstanding something, but I can't reproduce the second issue.

Tunous:
peek 2017-04-12 17-49

Tunous:
The bug with fab happens every time when title input field is empty.

Tunous:
Should hide expand button if can't comment (locked commenting, etc)

Tunous:
In a few days I'll have some more free time. I am going to try and finally finish this. 

-------

Another issue:
incorrect pop-up position
The further I scroll down in collapsed editor the higher user autocompletion pop-up appears.

-------

@maniac103 Did you make any progress on this pull request worth sharing before I commit anything?

maniac103:
@Tunous Nope, didn't have any time to work on it further :-(

Tunous:
Issue: Restoring state stopped working.

Tunous:
I solved the problem with FAB by hiding it when the input is empty. People shouldn't be able to save issues without title anyway :smile: 

State restoring works correctly in `CommitActivity` but is broken in `IssueActivity`. Apparently, it's a bug in how `IssueActivity` deals with its fragment. Instead of restoring it always creates a new one. @maniac103 I think that's something that you have to solve.

Other than that the only issue left is the one with weird scrolling https://github.com/slapperwan/gh4a/pull/536#issuecomment-293622580. I still didn't find what is causing it.

Tunous:
I've almost fixed the issue with the basic editor scrolling but there is one weird thing happening that I'm not sure how to resolve.

peek 2017-05-27 14-40

1. Enter some text so basic editor is scrollable
2. Scroll it up
3. Try to scroll content or toolbar

As you can see in the gif scrolling works nicely when the basic editor is scrolled to bottom. But if you scroll it a little bit up everything breaks.

I've tried debugging it and found that the cause is located in the scrolling behavior of AppBarLayout. Exactly it's this line:

https://github.com/android/platform_frameworks_support/blob/master/design/src/android/support/design/widget/AppBarLayout.java#L1432

I'm not sure what is the best way how to solve this. I don't know which view exactly implements this behavior and how to override it to not cause any other problems.

___

From other changes, I replaced the MarkdownUtils class with my new library. It can now be used in other projects too. I improved it a lot and added some unit test to be sure that everything works correctly. Source: https://github.com/Tunous/MarkdownEdit

Tunous:
I have fixed the remaining issue with scrolling. It's a bit hacky but at least now it works properly.

Now, this pull request **finally** reached a state where I don't see any obvious bugs or annoyances. Should be good for final reviews and merging.

app/src/main/java/com/gh4a/BaseActivity.java
```diff
@@ -586,9 +609,19 @@ private void setupSwipeToRefresh() {
...
+
+        mSwipeLayoutBehavior = new BottomSheetFixScrollingViewBehavior();
+        CoordinatorLayout.LayoutParams lp =
+                (CoordinatorLayout.LayoutParams) mSwipeLayout.getLayoutParams();
+        lp.setBehavior(mSwipeLayoutBehavior);
+
         updateSwipeToRefreshState();
     }
 
+    public void setBottomSheetFixEnabled(boolean enabled) {
```
maniac103: This is about the only part of the whole diff that I'm unhappy with, because this method is a bit unclear: from reading the code, it's not obvious what the 'bottom sheet fix' is and why it's needed.

I think I'd prefer a method that instantiates the swipe layout behavior instead, returning a normal ScrollingViewBehavior by default and a BottomSheetFixScrollingViewBehavior (BTW, name suggestion: 'BottomSheetCompatibleScrollingViewBehavior'?) for the activities that host the bottom sheet fragment.

Tunous:
Added 3 lost commits and implemented the suggested change.

maniac103:
Rebased to master again. Thanks for adding the missing commits!

Remaining open items that I can see:
- [ ] Selecting a commit (line) comment and choosing 'Add comment' copies the comment text we want to reply to into the edit box. Same happens with the 'Reply' action for PR comments.
- [ ] In the above scenario, the bottom sheet doesn't open above the IME. One has to close and reopen the IME to get the sheet to move up
- [ ] The 'Add comment' and 'Reply' editors should probably use a 'send' icon instead of the 'save' one.

Tunous:
This issue with keyboard is weird. Maybe it has something to do with the fact that advanced layout is inflated after a delay and first focus tries to target basic editor.

Tunous:
Committed a workaround for the keyboard issue which disables its automatic opening. Not the best option but that's better than keyboard which covers editor. I can't find any other solutions...

Tunous:
New issue:

- [ ] Fling scroll gesture doesn't work in advanced editor.

Tunous:
Rebased onto master

sharkykh:
Looks great!

Some constructive comments:

1. Something I wasn't able to record:
   _(do you have a suggestion for a good gif screen recorder, in high-res and high FPS?)_
   The comment sheet on pull requests "jumps" a bit when closed.
   When I mean is, it stops for a (half-a-)second in [this state](https://user-images.githubusercontent.com/10238474/29058165-90a12a04-7c17-11e7-929d-17983ccefa15.png), before jumping to [the final state](https://user-images.githubusercontent.com/10238474/29058170-9304b310-7c17-11e7-9c5c-5a7cba6ca17f.png).

2. Is the placeholder text [supposed to be in 2 rows like that](https://user-images.githubusercontent.com/10238474/29058290-37562f16-7c18-11e7-90c3-f48912e1b9ac.jpg)?
I mean, it fixes itself when you expand the comment sheet, but it looks a bit weird, since it doesn't happen on issues.

3. Love the Preview tab! I tested it with the contents of [test.mdown](https://github.com/slapperwan/gh4a/blob/master/test.mdown).
Are you planning on implementing [the auto-links](https://user-images.githubusercontent.com/10238474/29058415-e30c55c4-7c18-11e7-81b1-c093034924d1.jpg) as well?
I know they don't show up on the `.m[ark]d[own]` files, but they do in issues and PRs,
for example:
* SHA: 77604a345985facf0c1f82c8b8f0930c692d33cd
* User@SHA ref: sharkykh@5e3ab0b316f29f5db57a6900a93a7471fe1b51c5
* User/Project@SHA: sharkykh/gh4a@5e3ab0b316f29f5db57a6900a93a7471fe1b51c5
* \#Num: #536
* User/#Num: maniac103#1
* User/Project#Num: sharkykh/TestRepository#1

Tunous:
> (do you have a suggestion for a good gif screen recorder for the phone, in high-res and high FPS?)

I'm using https://github.com/vijai1996/screenrecorder and it's working well.

> Is the placeholder text supposed to be in 2 rows like that?

This text just doesn't fit on your device so it wraps. You should be able to scroll up to see the whole message. I don't think we can fix that other than making the text message shorter, but that still could wrap on even smaller devices.

> The comment sheet on pull requests "jumps" a bit when closed.
When I mean is, it stops for a (half-a-)second in this state, before jumping to the final state.

I guess that this one is caused by tabs bar disappearing. There is probably a nice way to also animate that change of visibility...

As for the markdown rendering I think that should be addressed in separate issues/pull requests (like [this one](https://github.com/slapperwan/gh4a/issues/530)) since it also affects viewing markdown files. This pull request doesn't need these changes to be merged.

maniac103:
Rebased onto master again. The fling in advanced editor not working is the only issue left? If that's the case, I think we should finally merge this PR and leave the fling fix for a separate commit.

Tunous:
Yep that's the last issue left. 
We should test it with replying to review comments now that it's rebased. Bottom sheet instead of the old dialog needs to be used here too.

maniac103:
I did the dialog -> bottom sheet conversion for review replies during the rebase, as otherwise it wouldn't even compile (dialog layout resource missing).

Tunous:
So it should be merged. I am using this branch for few months already and didn't find any more annoyances. 

Note that this pull request also updates SDK so the other SDK update pull request can be merged too.

app/src/main/java/com/gh4a/activities/DiffViewerActivity.java
```diff
@@ -453,9 +426,13 @@ protected void handleUrlLoad(Uri uri) {
         boolean isComment = idParam != null;
         long id = isComment ? Long.parseLong(idParam) : 0L;
 
-        CommentActionPopup p = new CommentActionPopup(id, line, lineText, leftLine, rightLine,
-                mLastTouchDown.x, mLastTouchDown.y, isComment, isRightLine);
-        p.show();
+        if (idParam == null) {
```
Tunous: Is this correct? Shouldn't the pop-up be displayed every time instead?

maniac103: Yes, I think so. Previously only the 'add comment' item applied, but now we also have 'share'.

Tunous:
Ok I went through all the modified files and didn't find any other obvious issues.

Tunous:
Should I merge this?

maniac103:
I can do it as well :)
@slapperwan I think with this feature, next version should be called 4.2 ;-)

PR #547, opened 2017-03-23
Fast scroller

### LINKED ISSUE #233: Add fast scrolling again
You switched layout of issue content to fast scrolling some time ago. It is not working anymore. Any reason why? If has long content you need to scroll a long time.

Tunous:
Closes #233.

### DIFF
--- a/res/layout/generic_list.xml
+++ b/res/layout/generic_list.xml
@@ -1,7 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<com.gh4a.widget.ContextMenuAwareRecyclerView
+<FrameLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/list"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    android:scrollbars="vertical" />
+    android:layout_height="match_parent">
+
+    <com.gh4a.widget.ContextMenuAwareRecyclerView
+        android:id="@+id/list"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+
+    <com.pluscubed.recyclerfastscroll.RecyclerFastScroller
+        android:id="@+id/fast_scroller"
+        android:layout_width="wrap_content"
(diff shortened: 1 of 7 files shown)


src/com/gh4a/fragment/LoadingListFragmentBase.java
```diff
@@ -54,6 +57,13 @@ protected View onCreateContentView(LayoutInflater inflater, ViewGroup parent) {
                     UiUtils.resolveDrawable(getActivity(), R.attr.listBackground));
         }
 
+        mFastScroller = (RecyclerFastScroller) view.findViewById(R.id.fast_scroller);
+        mFastScroller.attachRecyclerView(mRecyclerView);
+
+        int color = UiUtils.resolveColor(view.getContext(), R.attr.colorControlNormal);
+        color = Color.argb(93, Color.red(color), Color.green(color), Color.blue(color));
```
maniac103: Is there a specific reason not to use an XML defined color here? I think I'd prefer that.

res/layout/generic_list.xml
```diff
@@ -1,7 +1,21 @@
...
+    <com.gh4a.widget.ContextMenuAwareRecyclerView
+        android:id="@+id/list"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+
+    <com.pluscubed.recyclerfastscroll.RecyclerFastScroller
+        android:id="@+id/fast_scroller"
+        android:layout_width="wrap_content"
+        android:layout_height="match_parent"
+        android:layout_gravity="end"
```
maniac103: right or right|end

(min API is < 17 and we don't have RTL language translations anyway)

src/com/gh4a/fragment/LoadingListFragmentBase.java
Tunous: Changed. I did it like this because I tried to replicate what *(I think)* happens with original scrollbar thumb where there is a tint of another color that changes transparency. But you are right. Setting a color in a regular way is better ;)

res/layout/generic_list.xml
Tunous: Fixed.

PR #554, opened 2017-03-28
Open issue after creation

### LINKED ISSUE #458: Adding new issue doesn't show it in the list immediately

### LINKED ISSUE #496: After adding new issue user should be redirected to it instead of the issues list

Tunous:
The refresh code is no longer needed as a list refresh will be automatically executed when the user closes the newly opened activity.

Closes #496 and fixes #458.

### DIFF
--- a/src/com/gh4a/activities/IssueListActivity.java
+++ b/src/com/gh4a/activities/IssueListActivity.java
@@ -20,7 +20,6 @@ import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
-import android.os.Handler;
 import android.os.PersistableBundle;
 import android.support.annotation.StringRes;
 import android.support.design.widget.CoordinatorLayout;
@@ -52,6 +51,7 @@ import com.gh4a.loader.MilestoneListLoader;
 import com.gh4a.loader.ProgressDialogLoaderCallbacks;
 import com.gh4a.utils.ApiHelpers;
 
+import org.eclipse.egit.github.core.Issue;
 import org.eclipse.egit.github.core.Label;
 import org.eclipse.egit.github.core.Milestone;
 import org.eclipse.egit.github.core.User;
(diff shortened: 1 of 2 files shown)


maniac103:
The latter doesn't hold true though when actually closing the new activity (back button instead of back arrow - the latter opens a new issue list activity).

Tunous:
What about the change in the above commit? It always refreshes the issues list even when the result was not set. It worked for me when I used back button.

src/com/gh4a/activities/IssueListActivity.java
```diff
@@ -333,11 +324,17 @@ protected boolean fragmentNeedsRefresh(Fragment object) {
...
         if (requestCode == REQUEST_ISSUE_CREATE) {
-            if (resultCode == Activity.RESULT_OK) {
-                // delay refresh for a bit, as
-                // - the fragment might not be created yet
-                // - Github seems to need some time to reflect new issues in their response
-                mHandler.postDelayed(mRefreshRunnable, 500);
+            if (resultCode == Activity.RESULT_OK && data.hasExtra("issue")) {
+                final Issue issue = (Issue) data.getSerializableExtra("issue");
+                // Delay click as the fragment might not be created yet
+                mHandler.post(new Runnable() {
```
maniac103: This delay seems a little unsafe. if mOpenFragment is null, we probably rather need to memorize the passed issue and carry out the onItemClick when we actually have the fragment reference. Or - probably preferable - just do the `startActivityForResult(IssueActivity.makeIntent(this, mRepoOwner, mRepoName, issue.getNumber())` ourselves.

src/com/gh4a/fragment/IssueListFragment.java
```diff
@@ -111,9 +110,7 @@ public void onItemClick(Issue issue) {
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (requestCode == REQUEST_ISSUE) {
-            if (resultCode == Activity.RESULT_OK) {
-                super.onRefresh();
-            }
```
maniac103: Won't this also cause a refresh when canceling the 'new issue' activity with the back button?

Tunous:
Changed it so it doesn't deal with open issues fragment at all. It instead launches issue activity with a new request and always performs a refresh of all fragments when returning from it.

Issue #150, opened 2014-01-21
News feed: allow switching account context


Gama11:
This would be really helpful, currently, everything seems to be in the same feed. Both the (desktop) website and the official GitHub app have this option.

maniac103:
Working on this. It's actually not too hard, only some styling/UX work left to do.

PR #564, opened 2017-04-01
Click targets and highlihts improvements

Tunous:
- Now possible to enter own user screen when clicking anywhere inside of the drawer title instead of just the avatar
- Search suggestions now display click highlight

### DIFF
--- a/res/layout/drawer_title_home.xml
+++ b/res/layout/drawer_title_home.xml
@@ -1,27 +1,35 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout
+<FrameLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:layout_gravity="center_vertical"
-    android:orientation="horizontal"
-    android:paddingTop="24dp"
     tools:background="?colorPrimary">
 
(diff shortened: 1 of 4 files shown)


maniac103:
+1 for the search suggestion changes, not sure about the nav drawer one though. Actually the current layout is more of a placeholder, and I eventually want it to behave like the Gmail drawer header, in which case the user's name would be removed (see #150). We might be able to kang some code from jruesga/rview for this.

Tunous:
Hmm, I'll leave the implementation of that header for you then as I'm not exactly sure how you want it to work.

You can just cherry-pick the search suggestion changes :)

maniac103:
I've implemented what I had in mind here: https://github.com/slapperwan/gh4a/tree/multi-account

It's mostly working, but the 'add new account' thing still needs a dialog that
- points the user to log out from Github in his browser beforehand and
- shows a indeterminate progress bar after starting the actual oauth process

Comments and patches are welcome ;)

Tunous:
> points the user to log out from Github in his browser beforehand

I found this: https://github.com/logout.

maniac103:
Yeah, sure, that's where the user needs to be pointed to. There's no way to automatically redirect him to the app after logging out, though, just as there's no way to enforcing a new login in the oauth flow :(

maniac103:
Pushed the suggestion background fix and the multi-account branch to master.

PR #574, opened 2017-04-11
Scroll to first unread comment when coming from notification

Tunous:
The API doesn't tell the id of that comment so instead I'm finding a comment based on the date when the issue was last viewed.


maniac103:
Thoughts about the last commit? I find it somewhat cleaner, even if it's on the edge to being overengineered.

Tunous:
Looks good.

PR #575, opened 2017-04-12
Drawer header improvements

Tunous:
Added click highlight again and fixed bottom padding.


maniac103:
The lack of click highlight is intentional here ... I used Gmail as role model, which doesn't have that highlight either.

Tunous:
I think it should have it as that's a clickable element. Can't see a reason why would it be better to not add it. Also Google is often inconsistent with their apps when following material design ;)

maniac103:
They are, but they're pretty consistent here. Out of the 11 Google apps I tried, only one (G+) had the click highlight for the header (no rule without exception), and for my taste it looked somewhat weird there. I'll give it a try tomorrow though.

PR #578, opened 2017-04-12
Add "Starred repositories" to navigation drawer

### LINKED ISSUE #388: Shortcut in sidebar to jump to my list of starred repos
I'd appreciate it if there were an additional quick access link in the sidebar to jump to the list of repositories I have starred.

Tunous:
Closes #388.

### DIFF
--- a/res/menu/home_nav_drawer.xml
+++ b/res/menu/home_nav_drawer.xml
@@ -41,6 +41,10 @@
             android:id="@+id/bookmarks"
             android:title="@string/bookmarks"
             android:icon="@drawable/icon_bookmark" />
+        <item
+            android:id="@+id/starred_repos"
+            android:title="@string/starred_repositories"
+            android:icon="@drawable/unstar" />
     </group>
     <group
         android:id="@+id/explore"
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -18,6 +18,7 @@
         <item>@string/my_gists</item>
         <item>@string/search</item>
         <item>@string/bookmarks</item>
+        <item>@string/starred_repositories</item>
         <item>@string/pub_timeline</item>
         <item>@string/trend</item>
         <item>@string/blog</item>
@@ -32,6 +33,7 @@
         <item>gists</item>
         <item>search</item>
         <item>bookmarks</item>
+        <item>stars</item>
         <item>timeline</item>
         <item>trends</item>
         <item>blog</item>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -103,6 +103,7 @@
     <string name="mentioned">Mentioned</string>
     <string name="participating">Participating</string>
     <string name="my_repositories">My repositories</string>
+    <string name="starred_repositories">Starred repositories</string>
     <string name="my_issues">My issues</string>
     <string name="my_pull_requests">My pull requests</string>
     <string name="my_gists">My gists</string>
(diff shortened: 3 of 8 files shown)


maniac103:
I'm not sure whether this really is a good idea, as this is doable as of now already. One can set 'my repositories' as home page, and the filter type is stored persistently, so when selecting 'starred' there, one would have the starred repositories as home page. Maybe we just should advertise this capability better?, e.g. using some one-time help cling?

If we're really going to add this, I'd prefer passing some initial filter type into the factory instead of special casing the 'starred' type.

Tunous:
This pull request was inspired by the implementation of starred repositories on GitHub page. Here when you go to your profile screen you can see that stars and repositories have separate entries. 

I personally wouldn't be against adding a hint showing users where to find starred repositories. There actually might be the problem as "My repositories" doesn't feel like a place where you would find watched or starred repositories. Maybe they should be separated instead?

Tunous:
> There actually might be the problem as "My repositories" doesn't feel like a place where you would find watched or starred repositories. Maybe they should be separated instead?

Now I'm convinced about this even more. I think having them located under My Repositories is not intuitive. 

Just now I was trying to search for my own repository and was confused why there are many repositories not owned by me. It turned out that I've previously viewed starred repos and the filter was remembered. 

There are 2 solutions I propose to fix this:

1. Move starred and also watched repositories entirely out of My Repositories section and place them in drawer. 
2. Rename My Repositories just to Repositories and get rid of auto-remembering current filter. 

I'm in favor with option 1.

@maniac103 thoughts?

Tommy-Geenexus:
@Tunous
I completely agree. The current solution is quite counter intuitive.
Putting them into the drawer would be a good thing to do.

maniac103:
I'm fine with option 1 I guess. The drawer might become a bit massive though, so we'll need to think about that (better grouping? allow hiding drawer items? not sure...)

Tunous:
I don't think it's that huge. Should be still ok with 2 more items.

As for the grouping perhaps add section for repositories: My Repositories, Starred Repositories, Watched Repositories and place that under group with Bookmarks. But I'm not sure if it's ok to move My Repositories entry...

TR4Android:
In my opinion a new item in the drawer would be too much, but a tabbed interface like in "My issues" and "My Pull Requests" would be nice and allow for quicker access to the starred and watched repositories.

sharkykh:
If you're worried about too many items in the drawer, I'd suggest having "Repositories" open up in the drawer as a submenu/group directly underneath the drawer item, if that's doable (don't know about the coding aspect of this)

For example:
```
Repositories    >

### clicking will toggle the display of the submenu

Repositories    v
   My Repositories
   Starred Repositories
   Watched Repositories
```

Tunous:
Check #624 for test tabs implementation.

maniac103:
[Suggestion](https://github.com/slapperwan/gh4a/issues/696#issue-256078829) by @smichel17 in #696:

> On the My repositories screen, the Starred and Watched filter categories feel a bit out of place, since they alone are not subsets of All.
>
> I think it would make sense to move Watched under Notifications, as its own tab (or possibly its own screen via a menu item). I seldom want to go look through the list of what I'm watching, unless I want to change that list -- in the context of notifications, and this mirrors the github interface, so we don't need to make people re-learn where it is.
>
> Then, Starred can go under Bookmarks, which serves basically the same purpose, anyway. There could either be two tabs, or they could be in the same list, with bookmarks appearing on top.

Moving starred under bookmarks makes sense, but the bookmarks page also allows bookmarking users ... not sure whether this really fits. Moving watched under notifications seems to make sense to me as well, but I'm only a rare user of notifications and don't use (explicit, as opposed to automatic) repo watching myself, so take my opinion with a grain of salt there ;-)

#695 also provides another idea about drawer organization.

Tunous:
I've implemented the part about moving watched repositories.

For stars I propose to rename entry in drawer to one of these:

1. "Bookmarks and Stars"
2. "Stars and Bookmarks"
3. "Bookmarsk & Stars"
4. "Stars & Bookmarks"

And move them here as a tab similar to what I did with watched repos.

Tunous:
Decided to merge #734 instead of this.

PR #589, opened 2017-04-22
Introduce flavors and api key abstraction for F-Droid [F-Droid 3/4]

thermatk:
Please generate the necessary api keys for a separate flavor and paste them into the ApiKeyHelper.java and it's done. Then I'll go and update the f-droid metadata build script :+1:


app/build.gradle
```diff
@@ -39,8 +45,20 @@ android {
...
+    productFlavors {
+        // includes proprietary libs
+        fabric {
+            buildConfigField "boolean", "FDROID", "false"
+        }
+
+        // only foss
+        fdroid {
+            buildConfigField "boolean", "FDROID", "true"
         }
```
maniac103: I think I'd prefer if the flavors would be called 'foss' and 'playstore'.

app/src/fabric/java/com/gh4a/FabricApplication.java
```diff
@@ -0,0 +1,17 @@
...
+
+import com.crashlytics.android.Crashlytics;
+
+import io.fabric.sdk.android.Fabric;
+
+public class FabricApplication {
+    public static void onCreate(Application app) {
+        Fabric.with(app, new Crashlytics());
+    }
+    public static void trackVisitedUrl(String url, int sNextUrlTrackingPosition) {
```
maniac103: int position

app/src/fdroid/java/com/gh4a/FabricApplication.java
```diff
@@ -0,0 +1,12 @@
+package com.gh4a;
+
+import android.app.Application;
+
+public class FabricApplication {
+    public static void onCreate(Application app) {
+
+    }
+    public static void trackVisitedUrl(String url, int sNextUrlTrackingPosition) {
```
maniac103: As above

app/src/main/java/com/gh4a/Gh4Application.java
```diff
@@ -136,9 +133,9 @@ public void onCreate() {
         prefs.registerOnSharedPreferenceChangeListener(this);
 
         boolean isDebuggable = (getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0;
-        sHasCrashlytics = !isDebuggable && !TextUtils.equals(Build.DEVICE, "sdk");
+        sHasCrashlytics = !isDebuggable && !TextUtils.equals(Build.DEVICE, "sdk") && !BuildConfig.FDROID;
```
maniac103: Why not move all this logic to the helper class as well? Just name it CrashReportingHelper instead of FabricApplication and move sNextUrlTrackingPosition and sHasCrashlytics there and then have this method there:

```
public static void init(Context context) {
    sHasCrashlytics = <...>
    if (sHasCrashlytics) {
        Fabric.with(context, new Crashlytics());
    }
}

File: app/src/main/java/com/gh4a/Gh4Application.java [hunk 5]
```diff
@@ -196,8 +193,7 @@ public void onConfigurationChanged(Configuration newConfig) {
     /* package */ static void trackVisitedUrl(String url) {
         synchronized (Gh4Application.class) {
             if (sHasCrashlytics) {
-                Crashlytics.setString("github-url-" + sNextUrlTrackingPosition, url);
-                Crashlytics.setInt("last-url-position", sNextUrlTrackingPosition);
+                FabricApplication.trackVisitedUrl(url, sNextUrlTrackingPosition);
```
maniac103: As above: Move all the logic into the helper class and only do
```
CrashReportingHelper.trackVisitedUrl(url);
```

here inside the synchronized block.

app/src/main/java/com/gh4a/utils/ApiKeyHelper.java
```diff
@@ -0,0 +1,24 @@
+package com.gh4a.utils;
+
+import com.gh4a.BuildConfig;
+
+public class ApiKeyHelper {
+    private static final String GITHUB_FDROID_CLIENT_ID = "SOMETEXT";
+    private static final String GITHUB_FDROID_SECRET = "SOMETEXT";
```
maniac103: Is there no way to achieve this with a pure gradle solution, that is, put the hardcoded values in build.gradle and use it depending on the flavor there?

Also, publishing the client secret allows other people to impersonate Octodroid, which doesn't seem like a good idea to me. Isn't there a way to put the additional properties file into the F-Droid metadata? From a quick glance over other app's metadata, it looks like patches can be supplied, so it should be possible to submit a patch adding the properties file into their repo.

thermatk: F-droid metadata is all open as well, all the other apps do it like that as well. I can get your keys from the apk in 5 minutes no matter how hard you obfuscate, so it's not a true thing to be concerned about :+1:

thermatk:
will apply changes when I come home

app/build.gradle
thermatk: done

app/src/fabric/java/com/gh4a/FabricApplication.java
thermatk: done

app/src/fdroid/java/com/gh4a/FabricApplication.java
thermatk: done

thermatk:
> Why not move all this logic to the helper class as well?

My idea behind this is to keep the flavor-specific files to the minimum. They should be as lean as they can, just to provide one with a way not to import potentially-non-existent dependencies into main logic classes. Moving more logic there can potentially make maintenance more complex
> API secrets

They are no real secrets even in production APKs, you can't really do much with them and people always have to explicitly give permission to the app to access and do something. Hiding them unconditionally for the foss flavor looks better to me then hardcoding them on the frontpage in gradle. My proposal is for you to generate a separate pair with the proper callback url, "OctoDroid FOSS" probably and paste them here. "foss" flavor will compile with them on f-droid servers, everyone else uses playstore config anyway :+1:

maniac103:
>> Why not move all this logic to the helper class as well?
> 
> My idea behind this is to keep the flavor-specific files to the minimum. They should be as lean as they can, just to provide one with a way not to import potentially-non-existent dependencies into main logic classes. Moving more logic there can potentially make maintenance more complex

I believe one should still keep logic that belongs together at one place. Doing so will make the code more understandable and thus easier to maintain.

About the API keys, it doesn't really matter to me whether they're actually secret or not - shipping the respective properties file in a patch in F-Droid simply is the most elegant solution to me; there is no real need to have an override in the Java code.

thermatk:
Please provide the patch with API keys to be applied

app/src/playstore/java/com/gh4a/utils/CrashReportingHelper.java
```diff
@@ -0,0 +1,35 @@
...
+import android.text.TextUtils;
+
+import com.crashlytics.android.Crashlytics;
+import com.gh4a.BuildConfig;
+
+import io.fabric.sdk.android.Fabric;
+
+public class CrashReportingHelper {
+    private static final int MAX_TRACKED_URLS = 5;
+    private static int position = 0;
```
maniac103: sNextUrlTrackingPosition

app/src/playstore/java/com/gh4a/utils/CrashReportingHelper.java
```diff
@@ -0,0 +1,35 @@
...
+    private static boolean hasCrashlytics;
```
maniac103: sHasCrashlytics

app/build.gradle
```diff
@@ -43,6 +49,18 @@ android {
             it.buildConfigField 'String', 'CLIENT_SECRET', ClientSecret
         }
     }
+
+    productFlavors {
+        // includes proprietary libs
+        playstore {
+            buildConfigField "boolean", "FOSS", "false"
```
maniac103: This field isn't required anymore, right?

thermatk:
./

thermatk:
When you merge, will you provide the keys or modify metadata yourself?

maniac103:
That question is for @slapperwan .

(Note to self: don't forget to squash on merge)

slapperwan:
@thermatk can I just provide you the API keys then you create the patch and apply to metadata?

thermatk:
Sure!

thermatk:
you have to merge it so I can move forward, you know :rofl:

maniac103:
Oops, I kinda missed that important step ;-)

PR #609, opened 2017-05-24
Make code search results scroll to matching fragments

### LINKED ISSUE #597: Scroll to highlight when coming from search result

Tunous:
The appearance of code search results was changed so it's now possible to click on individual matched code blocks. When doing so the opened file will be automatically scrolled to the matched block.

peek 2017-05-24 14-55

Closes #597.

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/CodeSearchAdapter.java
+++ b/app/src/main/java/com/gh4a/adapter/CodeSearchAdapter.java
@@ -4,7 +4,6 @@ import android.content.Context;
 import android.graphics.Typeface;
 import android.support.v7.widget.RecyclerView;
 import android.text.SpannableStringBuilder;
-import android.text.style.RelativeSizeSpan;
 import android.text.style.StyleSpan;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -20,8 +19,15 @@ import org.eclipse.egit.github.core.TextMatch;
 import java.util.List;
 
 public class CodeSearchAdapter extends RootAdapter<CodeSearchResult, CodeSearchAdapter.ViewHolder> {
-    public CodeSearchAdapter(Context context) {
+    public interface Callback {
+        void onSearchFragmentClick(CodeSearchResult result, int matchIndex);
+    }
+
+    private final Callback mCallback;
+
+    public CodeSearchAdapter(Context context, Callback callback) {
         super(context);
+        this.mCallback = callback;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/gh4a/utils/StringUtils.java
```diff
@@ -139,4 +141,30 @@ public static SpannableStringBuilder applyBoldTags(Context context,
...
+                if (j == 0 && !cLine.endsWith(fLine) ||
+                        j != 0 && j != matchLines.length - 1 && !cLine.equals(fLine) ||
+                        j == matchLines.length - 1 && !cLine.startsWith(fLine)) {
+                    isMatch = false;
+                    break;
+                }
+            }
+
+            if (isMatch) {
+                return new Pair<>(i + 1, i + matchLines.length);
```
maniac103: Return an int[2] here?

app/src/main/java/com/gh4a/utils/StringUtils.java
```diff
@@ -139,4 +141,30 @@ public static SpannableStringBuilder applyBoldTags(Context context,
...
+        String[] matchLines = match.split("\n");
+
+        for (int i = 0; i < lines.length - matchLines.length; i++) {
+            boolean isMatch = true;
+            for (int j = 0; j < matchLines.length; j++) {
+                String cLine = lines[i + j];
+                String fLine = matchLines[j];
+                if (j == 0 && !cLine.endsWith(fLine) ||
+                        j != 0 && j != matchLines.length - 1 && !cLine.equals(fLine) ||
+                        j == matchLines.length - 1 && !cLine.startsWith(fLine)) {
```
maniac103: Suggest parenthesis around the && statements.

app/src/main/java/com/gh4a/adapter/CodeSearchAdapter.java
```diff
@@ -55,28 +59,48 @@ protected void onBindViewHolder(ViewHolder holder, CodeSearchResult result) {
                         int start = item.getStartPos();
                         int end = item.getEdndPos();
```
maniac103: Now that I see this, we should probably fix this :)

Tunous: What do you mean?

maniac103: getEdndPos -> getEndPos ;)

Tunous: Ohh I see now. Also, I'm not sure if you are using the app right now, but if you do then replying looks to be broken: 
spectacle x14496

Tunous:
Updated.

app/src/main/java/com/gh4a/utils/StringUtils.java
```diff
@@ -139,4 +140,30 @@ public static SpannableStringBuilder applyBoldTags(Context context,
         }
         return ssb;
     }
+
+    @Nullable
+    public static int[] findMatchingLines(String input, String match) {
+        String[] lines = input.split("\n");
+        String[] matchLines = match.split("\n");
+
+        for (int i = 0; i < lines.length - matchLines.length; i++) {
```
maniac103: There's one issue in this loop: If input and match are exactly equal, the break condition in this line will cause no match to be found.

app/src/main/java/com/gh4a/utils/StringUtils.java
```diff
@@ -139,4 +140,30 @@ public static SpannableStringBuilder applyBoldTags(Context context,
...
+            boolean isMatch = true;
+            for (int j = 0; j < matchLines.length; j++) {
+                String cLine = lines[i + j];
+                String fLine = matchLines[j];
+                if ((j == 0 && !cLine.endsWith(fLine)) ||
+                        (j != 0 && j != matchLines.length - 1 && !cLine.equals(fLine)) ||
+                        (j == matchLines.length - 1 && !cLine.startsWith(fLine))) {
```
maniac103: Wouldn't this be easier to read:

```
boolean lineMatches =
        j == 0 ? cLine.endsWith(fLine) :
        j == matchLines.length - 1 ? cLine.startsWith(fLine) :
        cLine.equals(fLine);
if (!lineMatches) {
    isMatch = false;
    break;
}
```

Another naming suggestion: cLine -> inputLine, fLine -> matchLine

app/src/main/java/com/gh4a/utils/StringUtils.java
```diff
@@ -139,4 +140,30 @@ public static SpannableStringBuilder applyBoldTags(Context context,
...
+                    isMatch = false;
+                    break;
+                }
+            }
+
+            if (isMatch) {
+                return new int[] { i + 1, i + matchLines.length };
```
maniac103: Shouldn't this be `{i, i + matchLines.length - 1}`? A match in the first line currently yields {1, 1}, not {0, 0}.

maniac103:
Maybe an even simpler matching algorithm could be this (untested):

```
int pos = input.indexOf(match);
if (pos < 0) {
    return null;
}
int total = 0, count = 0;
for (String line : lines.split("\n")) {
    if (pos >= total && pos < total + line.length()) {
        return new int[] { count, count + match.split("\n").length };
    }
    total += line.length;
    count++;
}
```

Doing so would have the advantage of having only a single loop, not two nested ones.

Tunous:
It's a bit off. What about this:

```java
int pos = input.indexOf(match);
if (pos < 0) {
    return null;
}

int start = input.substring(0, pos).split("\n").length;
int end = start + match.split("\n").length - 1;
return new int[] { start, end };
```

Looks to be working so far.

maniac103:
Yeah, this one seems to be working.

PR #610, opened 2017-05-24
Notifications indicator

### LINKED ISSUE #599: FR: Badge for notifications
I'd very much like, if possible, to have a badge on the notifications bell icon in the drawer, indicating whenever I have unread notifications. Basically just like on the site (the blue circle).

Possibly even have the number of notifications in it, if you can fit that in.
Or maybe just show the number when the amount is a single digit (1-9).

For your consideration.
Thanks!

Tunous:
screenshot_1495615866

Closes #599.


Tunous:
Instead of the icon could also display a number of notifications but I personally like how that little dot looks.

app/src/main/java/com/gh4a/BaseActivity.java
```diff
@@ -96,6 +100,10 @@
     private View mLeftDrawerHeader;
     private View mRightDrawerHeader;
     private SupportMenuInflater mMenuInflater;
+    private ImageView mNotificationsIndicator;
+
+    private Drawable mTintedCircleIcon;
+    private Drawable mCircleIcon;
```
maniac103: My gut feeling is that all of the changes in this file should be in HomeActivity. None of this is relevant to any other activity.

maniac103:
Wouldn't it be more obvious to just use a different icon instead of the dot? I agree it looks nice, but it's not exactly intuitive.

sharkykh:
> Wouldn't it be more obvious to just use a different icon instead of the dot? I agree it looks nice, but it's not exactly intuitive.

" <B>NEW</B> "?

Tunous:
screenshot_1495711084

What about this?

maniac103:
Yeah, I like the dot better than this as well. Just to be clear, by 'different icon' I meant using a different icon on the left side of the text when there are present notifications, not replacing the dot with something else.

Tunous:
screenshot_1495881548

This does look nice but it won't be easy to notice it unless you know where to look.
What about combining this icon with the dot? Dot will draw user attention and changed icon will make it easier to see what the dot means.

sharkykh:
> What about combining this icon with the dot? Dot will draw user attention and changed icon will make it easier to see what the dot means.

@Tunous Like on Github. A bell icon all the time, a bell with a blue dot when there are unread notifications.
I think I suggested this in #599 as well.

Tunous:
I don't think it's possible to use colored icons in this drawer implementation... Notice how it automatically changes text and icon to green. Similarly, it colors everything to that gray color when not selected.

sharkykh:
@Tunous 
Oh, that's too bad...
If that's the case - a dot to the right of the drawer item is better, IMO.

Tunous:
Now in addition to the dot, the icon on the left side is also changed. Also moved the code as suggested.

screenshot_1496488644

maniac103:
Looks great. Just had some minor comments.

Tunous:
Updated.

PR #611, opened 2017-05-24
Make repository attributes scrollable

### LINKED ISSUE #596: Repository attribute text doesn't fit
The repo attributes for a private fork are too long, so the size attribute text overflows.

In this case (both private and fork), I think it'd be appropriate to remove the "Private" and "Fork" text to make everything fit.

Screenshot:

Tunous:
peek 2017-05-24 18-21

Closes #596.


Tunous:
This introduced a bug. Opening repository doesn't work if you click on the scrollable area.

PR #618, opened 2017-06-03
Render emoji using separate library where necessary

### LINKED ISSUE #584: Emoji in repo description not being rendered
If a repo has an emoji in it's description, the emoji isn't rendered. Would be cool if we saw the actual emoji instead of just the text value, not sure if it's possible to do anything about this though.

Examples:
- [nodejs/node](https://github.com/nodejs/node)
- [avajs/ava](https://github.com/avajs/ava)

Tunous:
Affected places:
- Commit messages
- Activity events
- Repository descriptions

Closes #584

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -79,6 +79,9 @@ dependencies {
     compile 'com.github.castorflex.smoothprogressbar:library:1.1.0'
     compile 'org.ccil.cowan.tagsoup:tagsoup:1.2.1'
     compile 'com.github.pluscubed:recycler-fast-scroll:0.3.2@aar'
+    compile('com.vdurmont:emoji-java:3.2.0') {
+        exclude group: 'org.json', module : 'json'
+    }
     playstoreCompile('com.crashlytics.sdk.android:crashlytics:2.6.5@aar') {
         transitive = true;
     }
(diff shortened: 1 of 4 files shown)


TR4Android:
You might want to wait until Google's official [EmojiCompat](https://developer.android.com/preview/features/emoji-compat.html) library is out of beta :wink:.

Tunous:
This library serves a different purpose. What I've implemented is replacing emoji codes like `:+1:` to Unicode which can be understood by Android. The library you linked just makes sure that all emoji Unicode works on older devices. It doesn't deal with colon codes (or how they are called).

TR4Android:
@Tunous Yes, sorry about that confusion. I realized this soon after typing my comment :smile:.

Tunous:
No problem! I really like contributing to applications I often use :)

And I just noticed that this change will _most likely_ also render emoji inside of code blocks. But I'm not sure if it's worth trying to parse that correctly, especially for multiline code blocks.

maniac103:
Couldn't we avoid that by doing the emoji parsing in the HTML conversion code for the EventAdapter case? There we know what parts of the rendered text are code blocks.

Tunous:
> Couldn't we avoid that by doing the emoji parsing in the HTML conversion code for the EventAdapter case? 

We could but I have no idea where to place emoji parsing in that conversion code...

maniac103:
Actually, I think the code blocks are a non-issue because they aren't rendered in the activity stream (we don't get HTML for those).

PR #621, opened 2017-06-05
Bookmarks reordering

### LINKED ISSUE #601: Bookmarks reordering

Tunous:
peek 2017-06-05 20-38

Might need some additional testing (especially the database scheme update code) but it already works nicely.

As you can see in the gif I also changed how bookmark removing works. Now you swipe items right/left to remove them instantly. And you get an option to undo that action via the displayed snackbar.

I'm not sure about the reordering code. Perhaps there is an easier way to make it work? Right now if you quickly drag up or down the reorder code can skip some items and instead of shifting them all up/down it replaces dragged item with the last item.

Bonus improvement is a click highlight which was missing :)

Closes #601


app/src/main/java/com/gh4a/db/BookmarksProvider.java
```diff
@@ -83,6 +102,14 @@ public static boolean hasBookmarked(Context context, String url) {
         return hasBookmarked;
     }
 
+    public static void reorderBookmark(Context context, long id, int orderId) {
+        ContentValues cv = new ContentValues();
+        cv.put(Columns.ORDER_ID, orderId);
+        context.getContentResolver().update(Columns.CONTENT_URI, cv,
```
maniac103: Can use ContentUris.withAppendedId(Columns.CONTENT_URI, id) here instead of doing the selection below.

Tunous: Changed.

maniac103:
The problem mentioned with reordering is probably caused by the cursor being reloaded after the order ID change, which invalidates the data set. I guess for this to really work reliably, we'll have to either

1. load the cursor contents into a list of POJOs, which we can manipulate at will and sync back to the DB later or
2. put changed order into a SparseArray (DB ID -> order position) which - if present - takes precedence over the cursor content.

1 is probably easy to implement, but I'm unsure about its performance with a huge number of bookmarks (is this a use case worth optimizing for?). 2 will probably be hard to get right, if at all possible.

Tunous:
Implemented different solution. Similar to 2nd but a bit simpler. Now it works correctly for me.

app/src/main/java/com/gh4a/fragment/BookmarkDragHelperCallback.java
```diff
@@ -0,0 +1,36 @@
+package com.gh4a.fragment;
```
maniac103: This should probably reside in a different package or (preferably?) in BookmarkListFragment itself.

app/src/main/java/com/gh4a/adapter/BookmarkAdapter.java
```diff
@@ -105,31 +133,52 @@ public int getItemCount() {
...
+        Uri uri = ContentUris.withAppendedId(BookmarksProvider.Columns.CONTENT_URI, id);
+        context.getContentResolver().delete(uri, null, null);
+
+        Snackbar.make(viewHolder.itemView, R.string.bookmark_removed, Snackbar.LENGTH_LONG)
+                .setAction(R.string.undo, new View.OnClickListener() {
+                    @Override
+                    public void onClick(View v) {
+                        BookmarksProvider.saveBookmark(context, name, type, url, extraData, false);
+                    }
+                }).show();
```
maniac103: Nit: the .show() should be on a new line.

app/src/main/java/com/gh4a/adapter/BookmarkAdapter.java
```diff
@@ -105,31 +133,52 @@ public int getItemCount() {
...
     }
 
-    static class ViewHolder extends RecyclerView.ViewHolder {
+    class ViewHolder extends RecyclerView.ViewHolder implements View.OnTouchListener {
```
maniac103: I think I'd prefer passing the interact listener into the ViewHolder's constructor over making this non-static.

Tunous:
Updated and rebased.

PR #622, opened 2017-06-05
Fix icon in README and add myself as contributor

Tunous:


maniac103:
That reminds me ... @slapperwan what do you think of adding @Tunous as collaborator?

slapperwan:
Aha...@maniac103 @Tunous  100% agreed!

Tunous:
Thanks :)

Issue #494, opened 2017-03-04
Share link to file line


Tunous:
Perhaps by pressing on the line number like GitHub does?

Bluscream:
Sounds good :)

PR #623, opened 2017-06-06
Line link sharing

Tunous:
This is an incomplete change which adds line link sharing _only_ for diffs. It's built on top of the existing popup for comments. I've improved it to use standard Android popup and added Share option to it for this new action. It also works for sharing links to comments.

image

Similar popup could be implemented for non-diff files. I didn't do it here as it needs some javascript code to make lines clickable and I'm not really sure what's a nice way to do this. Alternatively, for these files only line numbers could be clickable but similarly, I have no ideas how to add that...

#494

### DIFF
--- a/app/src/main/assets/codeutils.js
+++ b/app/src/main/assets/codeutils.js
@@ -9,6 +9,22 @@ window.applyLineWrapping = function(enabled) {
     s.textContent = enabled ? 'pre { white-space: pre-wrap }' : '';
 };
 
+window.addClickListeners = function() {
+    var pre = document.getElementById("content");
+    var ol = pre != null ? pre.getElementsByTagName("ol")[0] : null;
+    var lines = ol != null ? ol.getElementsByTagName("li") : null;
+    if (lines != null) {
+        function listener(target, line, event) {
+            if (event.target === target) {
+                NativeClient.onLineTouched(line);
+            }
+        }
+        for (i = 0; i < lines.length; i++) {
+            lines[i].addEventListener('click', listener.bind(null, lines[i], i + 1));
+        }
+    }
+};
+
 window.scrollToHighlight = function() {
     if (!window.highlightTop || !window.highlightBottom || !window.innerHeight) {
         return;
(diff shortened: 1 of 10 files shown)


maniac103:
I'd prefer not to refactor the menu here, as I already did it in the reactions branch. I guess I should merge that one into master ;)

maniac103:
Pushed the reactions branch to master.

Tunous:
Updated.

maniac103:
Text line sharing is going to be tough. Prettify generates the line numbers via a ::before CSS selector, which doesn't allow adding DOM elements such as links - and I'm not sure I really want to change prettify :-/ Github's web site renders code in a HTML table, which allows for linking the table cells.
We could probably make the whole line clickable (after prettify ran, iterate over all `<li>` elements, wrapping their content in `<a>` elements), but I'm not sure of the performance of doing that and it also may be a bit cumbersome to use?

maniac103:
Something like this might work: https://gist.github.com/maniac103/3f5d6203f47070f86e4766f8bc6a8f4e
It only triggers on the li tag, not the span that contains the text. It still triggers on whitespace to the right of the text though.

Tunous:
I think that triggering on whitespace is acceptable here. Not sure how you could fix that.

The x and y coordinates for the popup seem to be off though.

maniac103:
> The x and y coordinates for the popup seem to be off though.

That's probably because the top of the WebView is scrolled under the toolbar?

Tunous:
Idk maybe. Could instead use the code from DiffViewerActivity which tracks correct position.

maniac103:
Yeah, moving that code to WebViewerActivity is probably the better idea.

Tunous:
Do you want to finish this one? Or should I take your changes and continue on top of them.

maniac103:
If you got some free time, please go ahead. Mine is pretty limited at the moment :(

Tunous:
Done. _Note: First merge #634_

maniac103:
I think you need to remove the first commit (which was in #634) from this PR.

Tunous:
Done.

maniac103:
Thanks 👍

Tunous:
Btw, I noticed that there are few old branches that aren't used. Probably should clean up them as they just clutter the list.

PR #624, opened 2017-06-06
Repositories tabs

Tunous:
A test implementation of repository tabs as suggested in https://github.com/slapperwan/gh4a/pull/578#issuecomment-301339782.

screenshot_1496770057
screenshot_1496770053

Didn't finish as I'm not sure about this. In finished implementation filters/sort orders should update when changing pages. This could be confusing...

### DIFF
--- a/app/src/main/java/com/gh4a/activities/RepositoryListActivity.java
+++ b/app/src/main/java/com/gh4a/activities/RepositoryListActivity.java
@@ -56,7 +56,7 @@ public class RepositoryListActivity extends FragmentContainerActivity implements
         updateRightNavigationDrawer();
 
         ActionBar actionBar = getSupportActionBar();
-        actionBar.setTitle(R.string.user_pub_repos);
+        actionBar.setTitle(R.string.repositories);
         actionBar.setSubtitle(mUserLogin);
         actionBar.setDisplayHomeAsUpEnabled(true);
     }
--- a/app/src/main/java/com/gh4a/activities/home/HomeActivity.java
+++ b/app/src/main/java/com/gh4a/activities/home/HomeActivity.java
@@ -72,7 +72,7 @@ public class HomeActivity extends BasePagerActivity implements
     static {
         START_PAGE_MAPPING.put(R.id.news_feed, "newsfeed");
         START_PAGE_MAPPING.put(R.id.notifications, "notifications");
-        START_PAGE_MAPPING.put(R.id.my_repos, "repos");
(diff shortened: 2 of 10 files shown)


TR4Android:
@Tunous I like it, it's definitely an improvement over the current situation :+1:. But I'm the one who suggested it so I don't think my opinion really counts :wink:.

maniac103:
As for the confusion, I'd suggest keeping the sort order intact when switching between tabs (have a single sort order selection for all 3 tabs) and only hiding the filter options on the starred and watched tabs (as those have an implicit filter set).

As for the tabs themselves, I'm torn. On one hand I don't like 'tabifying' of stuff if not strictly needed, but on the other hand there's a significant amount of people who currently struggle to find the starred repositories, so this is certainly an improvement. The only other option that comes to my mind is adding multiple entries to the drawer, but the multiple tabs approach probably fits better into our current model (see e.g. the 'My gists' section), so it's fine, I guess :)

smichel17:
@maniac103 I generally agree with your sentiment about tabifying[1]. However, it's just a fact that there's a ton of information on github, and it's all got to fit into one app somehow[2]. My own suggestion for where to put this is in #696.

[1]: see FastHub for an example of tabifying gone wrong. Good lord, there's so many moving interface pieces -- while the functionality is great, it's so difficult to use, even with extra animations turned off.

[2]: Honestly, I wouldn't be upset to remove file browsing functionality except the readme and have a companion app for all the git stuff (which would work for any website and could also be used by LabCoat for GitLab, etc). It'd be hard to pull off well, though.

maniac103:
> [1]: see FastHub for an example of tabifying gone wrong. Good lord, there's so many moving interface pieces -- while the functionality is great, it's so difficult to use, even with extra animations turned off.

Agreed (I've outlined similar thoughts at https://github.com/slapperwan/gh4a/issues/681#issuecomment-327026113) - I want to avoid cramming too much stuff into the UI because I don't like the outcome seen in FastHub and Gitskarios.

> [2]: Honestly, I wouldn't be upset to remove file browsing functionality except the readme and have a companion app for all the git stuff (which would work for any website and could also be used by LabCoat for GitLab, etc). It'd be hard to pull off well, though.

That won't happen, simply because it doesn't fit the way I use the app myself ;-) I want Octodroid to do whatever meaningfully can be done 'on the go' and reviewing code (be it in PRs or the repo) is part of that. One of my most frequent use cases of the app is following code links posted in some IRC or Slack channels, and the code browsing is needed for that.

Tunous:
Decided to merge #734 instead of this.

PR #627, opened 2017-06-07
Add loading of all and participating notifications

Tunous:
Added as fragment tabs but maybe it is better to use filters in the right drawer. Switching to other tabs would be rare so there is no need to waste data by loading all the tabs when you enter notifications screen just to see unread notifications. Alternatively, maybe there is a way to not load data in other fragments until you switch to them?

### DIFF
--- a/app/src/main/java/com/gh4a/fragment/NotificationListFragment.java
+++ b/app/src/main/java/com/gh4a/fragment/NotificationListFragment.java
@@ -39,15 +39,20 @@ import java.util.Date;
 
 public class NotificationListFragment extends LoadingListFragmentBase implements
         RootAdapter.OnItemClickListener<NotificationHolder>,NotificationAdapter.OnNotificationActionCallback {
-    public static NotificationListFragment newInstance() {
-        return new NotificationListFragment();
+    public static NotificationListFragment newInstance(boolean all, boolean participating) {
+        NotificationListFragment fragment = new NotificationListFragment();
+        Bundle args = new Bundle();
+        args.putBoolean("all", all);
+        args.putBoolean("participating", participating);
+        fragment.setArguments(args);
+        return fragment;
     }
 
     private final LoaderCallbacks<NotificationListLoadResult> mNotificationsCallback =
             new LoaderCallbacks<NotificationListLoadResult>(this) {
         @Override
         protected Loader<LoaderResult<NotificationListLoadResult>> onCreateLoader() {
-            return new NotificationListLoader(getContext());
+            return new NotificationListLoader(getContext(), mAll, mParticipating);
         }
 
         @Override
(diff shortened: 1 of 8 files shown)


Tunous:
And I found a way to add lazy loading.

maniac103:
Is there really a compelling use case for the 'all' page? If not, we could get away with an action bar toggle button to select between 'all unread' and 'unread for participating only'.

Tunous:
Tabs was the first idea I got so I went for it. That toggle indeed sounds better. I'll add that.

Now I have a question about lazy tab loading. Should I use it for tabs in repositories screen?

maniac103:
What makes the repo tabs special there? From my POV, we either should do it for all tabs or for none ... not sure whether it's really worth it at this point.

Tunous:
Repo tabs would be special because that's where I encountered data usage limit. But now that I think about it, that's probably because I was refreshing/restarting the app too often. I guess you are right.

Tunous:
> Is there really a compelling use case for the 'all' page?

I misunderstood you here. Yes, there is a use case for that. I needed that myself to check old notifications that I (accidentally) marked as read.

But I added submenu instead of toggle so you can select any of these 3 options.

___

Now I think I found a bug with the loader. When calling `onRefresh` it doesn't create a new loader and load new data but use the one that was previously here. Is this the method I should be using? If yes then need to fix that issue somehow.

maniac103:
No, you'll need to call destroyLoader() and call initLoader() again if you need a new loader. onRefresh() just pretends a content change to force the loader to reload the data.

Tunous:
Got it. Updated.

app/src/main/java/com/gh4a/fragment/NotificationListFragment.java
```diff
@@ -192,11 +186,40 @@ public void onClick(DialogInterface dialog, int which) {
...
+                mParticipating = false;
+                item.setChecked(true);
+                reloadNotification();
+                return true;
+            case R.id.notification_filter_participating:
+                mAll = false;
+                mParticipating = true;
+                item.setChecked(true);
+                reloadNotification();
+                return true;
```
maniac103: This can probably be shortened to:

```
case R.id.notification_filter_unread:
case R.id.notification_filter_all:
case R.id.notification_filter_participating:
    mAll = itemId == R.id.notification_filter_all;
    mParticipating = itemId == R.id.notification_filter_participating;
    item.setChecked(true);
    reloadNotifications();
    return true;
```

Tunous: Done!

app/src/main/java/com/gh4a/fragment/NotificationListFragment.java
```diff
@@ -40,14 +40,17 @@
 public class NotificationListFragment extends LoadingListFragmentBase implements
         RootAdapter.OnItemClickListener<NotificationHolder>,NotificationAdapter.OnNotificationActionCallback {
     public static NotificationListFragment newInstance() {
-        return new NotificationListFragment();
+        NotificationListFragment fragment = new NotificationListFragment();
+        Bundle args = new Bundle();
+        fragment.setArguments(args);
+        return fragment;
```
maniac103: This change isn't really needed, right?

Tunous: Yeah it's not needed. I didn't notice that.

Tunous: Updated.

PR #635, opened 2017-06-15
Directly open and highlight comment from news feed

Tunous:
I feel like now all of these load tasks should be moved out of `BrowseFilter` to separate file(s).


maniac103:
👍 for moving this code out into a separate package ('resolver', 'intentlookup' or something among those lines).

PR #657, opened 2017-08-09
Add open in browser action to release list and release info screens

### LINKED ISSUE #633: Open in browser action for releases
For individual releases and also list of releases.

Tunous:
I've decided to set showAsAction to "never" as opening in browser is not a common action and I prefer the more minimal look of the toolbar when it displays only 3-dots icon.

Closes #633


maniac103:
:+1: for the release part of this, but is this really needed for the release list?

Tunous:
List in browser includes entries for non release tags.

maniac103:
We have such a list as well though (in the form of branch selection), and it should be relatively easy for us to add the tags to the release list as well.

Tunous:
Ok, removed it from release list screen.

PR #658, opened 2017-08-09
Update SDK to 26

### LINKED ISSUE #591: Laggy scroll
First off, great app!

I was looking at a user profile, and it seems to be somewhat laggy/jerky when scrolling. I think it might have to do with hiding the top fragment when scrolling down. 

Example, flicking to scroll won't work, it will just hide the top frightened and then sit there. Subsequent scrolling is fine.

Tunous:
Closes #591

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -15,6 +15,7 @@ repositories {
     mavenCentral()
     maven { url 'https://maven.fabric.io/public' }
     maven { url "https://jitpack.io" }
+    maven { url "https://maven.google.com" }
 }
 
 apply plugin: 'com.android.application'
@@ -27,11 +28,11 @@ android {
     signingConfigs {
         playStore
     }
-    compileSdkVersion 25
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 26
+    buildToolsVersion '26.0.1'
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 25
+        targetSdkVersion 26
         versionCode 53
         versionName "4.1.2"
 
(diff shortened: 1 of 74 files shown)


app/build.gradle
```diff
@@ -27,11 +28,11 @@ android {
...
         playStore
     }
-    compileSdkVersion 24
-    buildToolsVersion '25.0.2'
+    compileSdkVersion 26
+    buildToolsVersion '26.0.1'
     defaultConfig {
         minSdkVersion 16
-        targetSdkVersion 23
+        targetSdkVersion 26
```
maniac103: Are we sure this won't have any side effects? Updating the target SDK version disables compatibility behavior for behavior changes in newer Android versions after all. Glancing over the behavior changes document I don't see any obvious breakage for us, but I'd like us to try first nevertheless ;-)

Tunous: Right, I've previously looked at this when updating SDK  to 25 for 
 comment bottom sheet. See https://github.com/slapperwan/gh4a/pull/536#issuecomment-290664355 to see my findings. In short, the biggest potential issue is where we pass a large amount of data through Intents as they have a hard limit now. But I don't have ideas how to fix that one....

I'll take another look later and also check for changes to 26 which wasn't released when I tested previously.

Tunous:
Rebased onto master and reviewed [behavior changes](https://developer.android.com/about/versions/oreo/android-8.0-changes.html) and [migration guide](https://developer.android.com/about/versions/oreo/android-8.0-migration.html) for Android 8.0. Also tested on 8.0 emulator. Didn't find any issues.

maniac103:
Thanks for checking!

sharkykh:
Should the README section on how to build OctoDroid also be updated? (v24 => v26)

Tunous:
Yeah, I'll fix that

PR #665, opened 2017-08-21
Ability to delete branch after merge of PR

### LINKED ISSUE #329: [Feature Request] Ability to delete branch after merge of PR
After I merge a PR in to the branch, there should be an option for me to delete that source branch if it is not protected.

Tunous:
Closes #329


app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -520,6 +573,86 @@ private void handlePullRequestUpdate() {
...
+            return new RestoreBranchTask();
+        }
+
+        @Override
+        protected Reference run() throws Exception {
+            DataService dataService =
+                    (DataService) Gh4Application.get().getService(Gh4Application.DATA_SERVICE);
+
+            PullRequestMarker head = mPullRequest.getHead();
+            String owner = head.getRepo().getOwner().getLogin();
```
maniac103: getOwner() (and/or getRepo()?) might return null if the user left Github since the PR creation. Not sure how to test that though :-/

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -167,6 +173,22 @@ protected void onResultReady(Boolean result) {
         }
     };
 
+    private final LoaderCallbacks<Reference> mHeadReferenceCallback = new LoaderCallbacks<Reference>(this) {
```
maniac103: I wonder whether we could/should do this after loading the PR so we can re-use the loaded PR data and avoid calling getPullRequest() twice. We could still show the options menu and just hide the item until the ref is loaded.

github-api/src/main/java/org/eclipse/egit/github/core/service/DataService.java
```diff
@@ -40,6 +24,24 @@
...
+import java.util.Map;
+
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_BLOBS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_COMMITS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_GIT;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_REFS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_REPOS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_TAGS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_TREES;
```
maniac103: Was the reordering really required? ;-)

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
Tunous: Updated.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
Tunous: We could hide this action from the menu during onCreateOptionsMenu it if these are null. Would that work?

PR #676, opened 2017-08-29
Extract BrowseFilter load tasks to new package

Tunous:


maniac103:
Why not also move BrowseFilter in there?

Tunous:
Updated.

PR #687, opened 2017-09-05
Refactor link parsing

Tunous:
I've moved the parsing algorithm out of BrowseFilter to a separate class
so it was possible to introduce unit tests. There are now 68 tests which
result in 100% test coverage of the new LinkParser.

Only behavior change in the app introduced by this commit is that
"https:github.com" url now opens the main application screen instead of
a browser.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -86,6 +86,8 @@ dependencies {
     playstoreCompile('com.crashlytics.sdk.android:crashlytics:2.6.5@aar') {
         transitive = true;
     }
+    testCompile 'junit:junit:4.12'
+    testCompile 'org.robolectric:robolectric:3.4.2'
 }
 
 def Properties props = new Properties()
--- a/app/src/main/java/com/gh4a/resolver/BrowseFilter.java
+++ b/app/src/main/java/com/gh4a/resolver/BrowseFilter.java
@@ -5,32 +5,10 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 import android.support.v7.app.AppCompatActivity;
-import android.text.TextUtils;
 
 import com.gh4a.Gh4Application;
 import com.gh4a.R;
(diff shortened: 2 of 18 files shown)


maniac103:
Great 👍 

> Only behavior change in the app introduced by this commit is that
"https:github.com" url now opens the main application screen instead of
a browser.

Why is that? AFAICT the github.com URL would fall into parseUserLink(), and UserActivity.makeIntent() should return null if a null user name is passed in. What am I missing here?

Tunous:
https://github.com/slapperwan/gh4a/pull/687/files#diff-b312564cad4a5b5bc615685263735d92R60

I added that redirection. If the user really wants to open browser they could just go to it directly instead of clicking on links. That's why I assumed it would be wise to handle this case to redirect to app too. I don't see that there is anything we lose by doing so. 

I can revert this if there are reasons to not handle it that I'm not aware of.

maniac103:
> If the user really wants to open browser they could just go to it directly instead of clicking on links.

Not sure I understand? What do you mean by 'go to it directly'? If he set Octodroid to handle links, he'll have no chance of opening the browser except opening it manually and typing in the URL manually.

Tunous:
> opening it manually and typing in the URL manually

If the user really wants to open browser they could just go to **the browser** directly **and open GitHub here** instead of clicking on links.

maniac103:
Which seems to defeat the purpose of links ;-)
I'm not sure here. In general we only handle links that we have a direct viewer for, and the start page is not exactly a 1:1 mapping to the news feed. I think I'd prefer to take that particular one out.

Tunous:
Ok I'll revert that

Tunous:
Updated.

Tunous:
@maniac103 Anything missing here or is it ready for merge? I want this change so I can add some more link handling.

maniac103:
No, it's fine. Sorry, I missed merging it.

PR #691, opened 2017-09-06
Refreshing issues while in search mode clears input #686

MattiaPrimavera:
Removing the line solves the bug. Since `onRefresh()` method from `BaseActivity` seems called a lot, how should have test side-effects of removing the line ?

### DIFF
--- a/app/src/main/java/com/gh4a/BaseActivity.java
+++ b/app/src/main/java/com/gh4a/BaseActivity.java
@@ -440,7 +440,6 @@ public abstract class BaseActivity extends AppCompatActivity implements
 
     @Override
     public void onRefresh() {
-        supportInvalidateOptionsMenu();
         mSwipeLayout.setRefreshing(false);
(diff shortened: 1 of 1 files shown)


Tunous:
To be sure that there are no side effects a good idea would be to check all `onPrepareOptionsMenu` and `onCreateOptionsMenu` methods in activities that extend `BaseActivity`. If these have code that uses data which is reset during refresh, then they'll probably result in exceptions or incorrect items in menus. 

I guess that `PullRequestActivity` might be one of activities that would break. But I'm not 100% sure.

maniac103:
Yeah, PR activity is one example, issue activity is another one (which is why the patch won't help here - issue activity needs to invalidate the menu on refresh).
IMHO the better solution would be restoring the current search query to the SearchView after menu inflation.

Tunous:
Closing because this fix doesn't work correctly.

PR #697, opened 2017-09-07
swap repo menu item position

smichel17:
In my opinion, it's rather awkward that the menu item to open the right drawer is not the rightmost icon. So, I swapped them.


maniac103:
Yeah, that's a good point.

Issue #692, opened 2017-09-06
Reactive Functional Programming


MattiaPrimavera:
Hi guys, 

did you already consider moving to introduce the code base to the **reactive** world, using the some of the following **Rx** libraries ? 

- [RxJava2](https://github.com/ReactiveX/RxJava)
- [RxPermissions](https://github.com/tbruyelle/RxPermissions) 
- [RxActivityResult](https://github.com/VictorAlbertos/RxActivityResult) 
- [RxAndroid](https://github.com/ReactiveX/RxAndroid)
- [RxBindings](https://github.com/JakeWharton/RxBinding)

maniac103:
Yes, (at least) I did consider it. I don't think it provides enough additional value to justify rewriting the whole app + the bindings, though. I'm not even sure I'd use if I were to rewrite the app from scratch - the loader infrastructure simply seems sufficient for what we're doing.

In terms of functional programming, I'd consider making the app use a library for the Java8 stream API (e.g. https://github.com/aNNiMON/Lightweight-Stream-API) more important/useful.

@Tunous Any thoughts from your side?

Tunous:
Even though the current codebase is dated I find it very easy to work with. I don't see any huge issues in how the app performs that would justify the rewrite. Rx is very nice but there is no reason to use it.

MattiaPrimavera:
It's not issue-related but more an 'enhancement' purpose. Code would be clearified since all async operations would pass through the unique interface **Observer**, this would avoid wiring lots of **AsyncTasks** and **Loaders**, which split logic into multiple callbacks. 

**rxPermissions** and **reactivity result** eliminate the need to override **permissions** and **result** callbacks, does the developer does not always have to move among callbacks to follow the application flow,

maniac103:
> It's not issue-related but more an 'enhancement' purpose. Code would be clearified since all async operations would pass through the unique interface Observer, this would avoid wiring lots of AsyncTasks and Loaders, which split logic into multiple callbacks.

There is no logic split here. The Loader subclasses are the equivalent of `Observable`, while the loader callbacks are the equivalent of `Observer`. The AsyncTasks for saving stuff (which aren't related to the loaders) are a combination of both `Observer` and `Observable`, so if anything, we'd need to split logic there.

[...]

PR #698, opened 2017-09-08
Repository fragment rx refactor

MattiaPrimavera:
Newly **introduced** : 
- `rxJava2`
- `rxAndroid`
- `retrolambda` 

**removed** classes : 
- `FillReadmeTask`
- `ReadmeLoader` 

`FillReadmeTask` has been replaced by asynchronous operation concatenation from `flatMap` operator (`rxJava2`) while `ReadmeLoader` has been removed in favour of `RepositoryService`. 

The newly created `service` package is meant to gather `Service` classes, each one wrapping `data loading logic` that is currently contained into `Loaders`. This would have the advantage of reducing the number of `Loader` and `AsyncTask` classes used, plus it would `group` asynchronous data fetch processes by `Fragment`, instead of by `Data`. 

This would make the program logic easier to follow in my opinion, since `background services` populating the `view` would be all declared in one `Service` class, instead of multiple `Loader` and `AsyncTask` classes.


maniac103:
I don't necessarily agree with the sentiment that observables should be all put into a single class. Both cases in that example are solely used by RepositoryFragment, so why put that code elsewhere?
Also, did you try what happens when you rotate the device? ;-) The data reload that's going to happen with your solution needs to be avoided. Also, I think you forgot to convert onRefresh(), and I wonder whether there's a way to abstract away the error handling that former lived in the loader base class.

Why do you use flatMap() instead of doOnNext() in the readme case? Given no data transformation actually happens, this seems a little pointless.

Anyway, this seems like an interesting exercise. Let's see how it'll look like when all features of the Loader solution are in so we have a complete view on this.

maniac103:
And BTW, is RetroLambda really needed? IIRC one can compile Android with Java 1.8 lately.

maniac103:
Actually, I think the API bindings should be converted to return Observables, then we can use map() to do any custom transformations.

MattiaPrimavera:
> Edit: Scratch that, your service combines what previously was in the loader package. In the pure Java spirit[TM] this class would better be called RepositoryObservableFactory though ;-) 

Exactly, agreed on `name`, maybe something containing `Repository` would effectively be more appropriate since we're talking about `data fetch` methods. 

> Also, did you try what happens when you rotate the device? ;-) The data reload that's going to happen with your solution needs to be avoided. 

Nope, I didn't, as I said I just wanted to show you guys the idea and see what you thought before putting more time into this. But **rxJava** has a nice way to handle that ;) 

But yes, I knew I had to test **orientation change**, but thanks for the `onRefresh` hint, that I think I forgot :P 

> Why do you use flatMap() instead of doOnNext() in the readme case? Given no data transformation actually happens, this seems a little pointless. 

Well said ... gonna check

> And BTW, is RetroLambda really needed? IIRC one can compile Android apps with Java 1.8 lately.

Mmm I don't know actually, would be nice for it not to be mandatory although


> Actually, I think the API bindings should be converted to return Observables, then we can use map() to do any custom transformations.

could you detail this one ?

MattiaPrimavera:
> I wonder whether there's a way to abstract away the error handling that former lived in the loader base class. 

Do you mean this one ? 

```java
    } catch (RequestException e) {
           /* don't spam logcat with 404 errors, those are normal */
            if (e.getStatus() != 404) {
                throw e;
            }
        }
```

If yes, that can already be removed since the error handler is `empty`, I think if the method throws an Exception that we'll be catched by `doOnError` method ... 

But I'd probably go for the definition of a `Transformer` assembling all `error handling` for network calls in a single spot.

maniac103:
>> I wonder whether there's a way to abstract away the error handling that former lived in the loader base class.
> Do you mean this one ?

Nope, I mean https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/loader/LoaderCallbacks.java#L31-L34
The snippet you quoted is just for ignoring particular exceptions.

MattiaPrimavera:
Ok, my answer stays the same, that's to say that those errors as well should already be propagated to the `doOnError` method

MattiaPrimavera:
We wouldn't ... we would write something like 

```java
Observable<Data> data = Repository.getService()
  .compose(RxTools.handleErrors(view))
```

Reference to the view if we wanna show errors ...

MattiaPrimavera:
**Orientation changes** seems now correctly handled, **Observable** instance is indeed **cache** and persists to **activity recreation** through the use of **Singleton Pattern** 

`onRefresh` method should instead perform a new `network` request right ?

MattiaPrimavera:
In the meanwhile can I move to this to a switch statement or there's particular code styling conventions to fit ? 

```java
        if (id == R.id.cell_pull_requests) {
            intent = IssueListActivity.makeIntent(getActivity(), owner, name, true);
        } else if (id == R.id.tv_contributors_label) {
            intent = ContributorListActivity.makeIntent(getActivity(), owner, name);
        } else if (id == R.id.tv_collaborators_label) {
            intent = CollaboratorListActivity.makeIntent(getActivity(), owner, name);
        } else if (id == R.id.cell_issues) {
            intent = IssueListActivity.makeIntent(getActivity(), owner, name);
        } else if (id == R.id.cell_stargazers) {
            intent = WatcherListActivity.makeIntent(getActivity(), owner, name);
        } else if (id == R.id.cell_forks) {
            intent = ForkListActivity.makeIntent(getActivity(), owner, name);
        } else if (id == R.id.tv_wiki_label) {
            intent = WikiListActivity.makeIntent(getActivity(), owner, name, null);
        } else if (id == R.id.tv_downloads_label) {
            intent = DownloadsActivity.makeIntent(getActivity(), owner, name);
```

Little bit tiny gain in readability ... 

```java
        Context c = getActivity();
        switch (id) {
            case R.id.cell_pull_requests:
                intent = IssueListActivity.makeIntent(c, owner, name, true); break;
            case R.id.tv_contributors_label:
                intent = ContributorListActivity.makeIntent(c, owner, name); break;
            case R.id.tv_collaborators_label:
                intent = CollaboratorListActivity.makeIntent(c, owner, name); break;
            case R.id.cell_issues:
                intent = IssueListActivity.makeIntent(c, owner, name); break;
            case R.id.cell_stargazers:
                intent = WatcherListActivity.makeIntent(c, owner, name); break;
            case R.id.cell_forks:
                intent = ForkListActivity.makeIntent(c, owner, name); break;
            case R.id.tv_wiki_label:
                intent = WikiListActivity.makeIntent(c, owner, name, null); break;
            case R.id.tv_downloads_label:
                intent = DownloadsActivity.makeIntent(c, owner, name); break;
            case R.id.tv_releases_label:
                intent = ReleaseListActivity.makeIntent(c, owner, name); break;
            default:
                if (view.getTag() instanceof Repository)
                    intent = RepositoryActivity.makeIntent(c, (Repository) view.getTag());
                break;
```

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -72,44 +67,13 @@ public static RepositoryFragment newInstance(Repository repository, String ref)
...
-            tvPullRequestsCountView.setText(String.valueOf(result));
-        }
-    };
+    private Observable<String> mLoadReadme;
+    private Disposable[] mSubscriptions;
 
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        mSubscriptions = new Disposable[2];
```
Tunous: Maybe it would be better to use some sort of map? That way we would avoid potential out of bounds exceptions when someone adds new subscription but doesn't update that declaration.

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -125,6 +89,8 @@ public void onDestroyView() {
         super.onDestroyView();
         mImageGetter.destroy();
         mImageGetter = null;
+
+        for(Disposable d : mSubscriptions) if(d != null) d.dispose(); // clean up rx
```
Tunous: Try to avoid one liners. It's better to always add braces to loops, ifs, etc. Also add space before opening brace, and the comment is useless 🙂

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,80 @@
...
+import org.eclipse.egit.github.core.service.IssueService;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Locale;
+
+import io.reactivex.Observable;
+
+public class RepositoryService {
+    public static Observable<String> loadReadme = null;
```
Tunous: I would prefer if this was set to private. Instead add an additional method for resetting it to null. That way we'll avoid null pointer exceptions when someone tries to use this subscription directly instead of method.

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -167,8 +174,10 @@ public void onPause() {
     public void setRef(String ref) {
         mRef = ref;
         getArguments().putString("ref", ref);
+
         // reload readme
-        getLoaderManager().restartLoader(0, null, mReadmeCallback);
+        mSubscriptions[0] = mLoadReadme.compose(RxTools.applySchedulers()).subscribe();
```
Tunous: I think there is an error. I don't see `mLoadReadme` being set anywhere.

Tunous:
> onRefresh method should instead perform a new network request right

Yes, refresh allows users to request for updated data (someone could update readme since they entered this screen) so it shouldn't use cached data. 

The change to switch statement is a good idea 👍

app/src/main/java/com/gh4a/service/RepositoryService.java
MattiaPrimavera: just gotta check how to undo `cache` operator on the observable, or I wont be able fetching data from the network again :S :P

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
MattiaPrimavera: It was temporary, but nice consideration. Ideally, in order to share disposables unsubscription to avoid memory leaks, this logic should be shared over all Activities and Fragments, since every one of them could make use of an Observable, there's also **CompoundSubscription** object in **rx** that may help ...

MattiaPrimavera:
Is that appropriate for **RepositoryFragment** to still inherit from **LoadingFragmentBase** ?

maniac103:
Well, if we switch to using Rx, we'll still need a base class to handle the subscriptions transparently. We don't want to manage them on a per-fragment basis, and I *think* we'd want to use one of the many RxLoader implementations available.

MattiaPrimavera:
> Need a base class

Totally agreed

MattiaPrimavera:
> Well, if we switch to using Rx, we'll still need a base class to handle the subscriptions transparently. We don't want to manage them on a per-fragment basis 

Just added a **BaseFragment** class including a **CompositeDisposable** object to which all fragments subscription will be added in order for them to be all unsubscribe when **onDestroyView** is called.

This way implies we'll have to **add** every **Fragment disposable** to the **Composite** one from the parent **BaseFragment** class

maniac103:
Try accessing two different repos. The second one will show the readme of the first one because your solution doesn't consider the method arguments after the first one.

MattiaPrimavera:
**emptyCache** because storing the result for a new subscriber is called **caching** in **Rx** (**.cache()** operator) 

Still I agree that it doesn't seem such an elegant solution for the moment, still clearer and more beautiful than loaders and async tasks although :P

MattiaPrimavera:
> proper solution to the rotation problem 

Is this a requirement for this **pull** to be merged ? 

Than how we'll we proceed on moving the rest of fragments and activities performing web calls to Rx? 

One pull per-fragment or one huge enormous single pull

maniac103:
> still clearer and more beautiful than loaders and async tasks although :P

I see we're in deep disagreement there.

>> proper solution to the rotation problem
> Is this a requirement for this pull to be merged ?

Yes, it's a pretty hard requirement for me. We already **do** have a properly working solution; using Rx is not a value by itself in my point of view.

Maybe we're in a misunderstanding, so let me clarify. First, to recap, the properties of the current loader solution:
- loader and data life cycle tied to fragment/activity life cycle (+)
- rotation handled transparently (+)
- transparent error handling - error view is shown if exception occurs (+)
- clunky syntax, loaders are referenced by ID (-)

And of the AsyncTasks for action handling:
- transparent error handling - shows SnackBar in case of exception (+)
- transparent retry handling via SnackBar (+)
- not tied to fragment/activity life cycle (-) [could likely be solved by base class using a Loader internally]
- somewhat inconvenient usage (-)

My understanding of the goal of this PR was to provide a solution that keeps the (+) properties of the current solution and remedies (all or some of) the (-) properties on a given example (here: RepositoryFragment) so we can compare the solutions on a feature complete basis - comparing solutions with half of the features being thrown out is not a fair/meaningful comparison.

Tunous:
> Is this a requirement for this pull to be merged ?

Yes, it is. I have the same opinion as @maniac103  which I [mentioned previously](https://github.com/slapperwan/gh4a/issues/692#issuecomment-327952786):

> As long as it **won't remove existing functionality** or heavily disturb in development of new functionality.

In that case, error handling is the existing functionallity.

MattiaPrimavera:
I agree with both of you on replacing with a better solution, keeping all the (+), removing some (-) and gaining extra functionalities, thanks @maniac103 for your really clear comment. 


Agreed on trying to compare the approaches as well, that s why I wanted to know what was exactly missing before the solution could be compared (I directly said merge sorry :P I'm a little into Rx as u understood :P).

I'll fill and try to detail missing points underlined by @maniac103 to see how we would handle this in **Rx** ;)

MattiaPrimavera:
Little summary of last commits: 

- I tried with `RxLoader` but ended up removing it 
- I finally `cache` Observables by default to handle orientation change, and created an `ObservableManager` to bind all Observables to the Activity lifecycle (avoids **memory leaks**) 
- data refresh needs now **emptying** the `cache` instead of **restaring** loaders 

What keeps staying in the `view` : 

```java
    @Override
    public void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);

        // .......

        this.loadRepository().subscribe();

        // .......
    }

    public Observable<Repository> loadRepository() {
        return RepositoryService.loadRepository(this, mRepoOwner, mRepoName)
                .compose(RxTools.applySchedulers())
                .doOnNext(repository -> {
                    Log.d("TEST", "repository doOnNext method called");
                    mRepository = (Repository) repository;
                    updateTitle();
                    invalidateTabs();
                    // Apply initial page selection first time the repo is loaded
                    if (mInitialPage >= PAGE_REPO_OVERVIEW && mInitialPage <= PAGE_ACTIVITY) {
                        getPager().setCurrentItem(mInitialPage);
                        mInitialPage = -1;
                    }
                    setContentShown(true);
                    supportInvalidateOptionsMenu();
                });
    }
```

One service looks like : 

```java
    public static Observable loadRepository(Context context, String repoOwner, String repoName) {
        Gh4Application app = (Gh4Application) context.getApplicationContext();
        return Observable.create(e -> {
            org.eclipse.egit.github.core.service.RepositoryService repoService = (org.eclipse.egit.github.core.service.RepositoryService)
            Gh4Application.get().getService(Gh4Application.REPO_SERVICE);
            e.onNext(repoService.getRepository(repoOwner, repoName));
            e.onComplete();
        })
        .cache()
        .compose(RxTools.handle(app, "loadRepository"));
    }
```

P.S. : `.cache()` could probably be moved into the RxTools.handle Transformer

The service would specify its `cache key`, here `loadRepository`, in order to be able clearing the cache for a specific service only. 

To clear the cache for a specific service : (here clears the cache for loadReadme service before calling the service) 

```java
RxTools.emptyCache(mApp, "loadReadme")
                .flatMap(o -> loadReadme())
                .subscribe(o -> {
                    Log.d("TEST", "Loading Readme subscribe reloadData called: " + o);
                });
```

In order to clean the entire cache : 
```java
RxTools.emptyCache(mApp)
                .flatMap(o -> loadReadme())
                .subscribe(o -> {
                    Log.d("TEST", "Loading Readme subscribe reloadData called: " + o);
                });
```

What do you think of this approach ?

Tunous:
I've tested it and it doesn't work correctly. Try opening one repository, going back and opening another. It'll load cached data from the previous repository. This needs to be solved.

MattiaPrimavera:
Thanks man, I'll take care of this. On **fragments** I still got to listen to their lyfecicle events ...

MattiaPrimavera:
Binding to Activity **lyfecicle** now includes **emptying the cache** whenever the activity is destroyed, with no Orientation change occurring

```java  
      @Override
        public void onActivityDestroyed(Activity activity) {
            if(subscriptions != null && subscriptions.isDisposed())
                subscriptions.dispose();

            if (activity.isFinishing()) { // Activity Destroyed, emptying cache
                emptyCache(null);
            } else {
                // Orientation change case
            }
        }
```

MattiaPrimavera:
Added **error handling**, with transparency, since it's attached to the main **RxTools.handle** or **RxTools.handleNoCache** Transformers, so no extra line added to the views code to have error handling on every network call ... 

**handle** did indeed need to be called anyway, in order to **factor** Schedulers definition

MattiaPrimavera:
@maniac103  

> Why do you use flatMap() instead of doOnNext() in the readme case? Given no data transformation actually happens, this seems a little pointless. 

Well, it was to share the **subscriber** logic within the **onNext** call, I should have probably created a **Subscriber** or smthg variable in order to share that

MattiaPrimavera:
Little question: Where was the error handling procedure here ? 

```java
    private class UpdateWatchTask extends BackgroundTask<Void> {
        public UpdateWatchTask() {
            super(RepositoryActivity.this);
        }

        @Override
        protected Void run() throws IOException {
            WatcherService watcherService = (WatcherService)
                    Gh4Application.get().getService(Gh4Application.WATCHER_SERVICE);
            RepositoryId repoId = new RepositoryId(mRepoOwner, mRepoName);
            if (mIsWatching) {
                watcherService.unwatch(repoId);
            } else {
                watcherService.watch(repoId);
            }
            return null;
        }

        @Override
        protected void onSuccess(Void result) {
            if (mIsWatching == null) {
                // user refreshed while the action was in progress
                return;
            }
            mIsWatching = !mIsWatching;
            supportInvalidateOptionsMenu();
        }
    }
```

Thanks

MattiaPrimavera:
on last commit : https://github.com/slapperwan/gh4a/pull/698/commits/b869ae5baa8b542a9d40f5dd1573362b979af556 

A few `AsyncTasks` moved to **Rx**: 
- AsyncTask classes removed -> All async operations unified through the `Observable` interface 
- errors handled transparently within `RxTools`, one `compose` instruction to attach both **schedulers** and **error handler**, plus I added a **positive message** in case of successful response ... shouldn't we ? Maybe displaying that for a lower time ? 
- One single layer handling network services for `IssueLabelListActivity` restrained into `IssueLabelService` file 
- only lines concerning the `view` stay into the Activity class file, the rest is either moved within the **Service** file if the logic is attached to the view, or to **RxTools** file, if the logic is **centralized / shared**

app/src/main/AndroidManifest.xml
```diff
@@ -72,7 +72,7 @@
             android:configChanges="orientation|screenSize" />
         <activity android:name=".activities.IssueActivity" android:exported="false" />
         <activity android:name=".activities.IssueEditActivity" android:exported="false" />
-        <activity android:name=".activities.IssueLabelListActivity" android:exported="false" />
+        <activity android:name=".activities.issues_label.IssueLabelListActivity" android:exported="false" />
```
maniac103: Please don't mix in refactorings like those package moves. They make reviewing harder; and I don't really want this package move either.

app/src/main/java/com/gh4a/activities/issues_label/IssueLabelListActivity.java
```diff
@@ -214,8 +207,10 @@ private void updateFabVisibility() {
 
     private final class EditActionMode implements ActionMode.Callback {
         private final IssueLabelAdapter.EditableLabel mLabel;
+        private final Activity mActivity;
 
-        public EditActionMode(IssueLabelAdapter.EditableLabel label) {
+        public EditActionMode(Activity activity, IssueLabelAdapter.EditableLabel label) {
+            mActivity = activity;
```
maniac103: Passing activity is unneeded here, the class isn't static, thus the activity can be accessed as IssueLabelListActivity.this.

maniac103:
> Binding to Activity lyfecicle now includes emptying the cache whenever the activity is destroyed, with no Orientation change occurring

It (semi-)binds to activity lifecycle, but doesn't bind to fragment lifecycle at all. Consider e.g. IssueActivity with a few reloads that each trigger a fragment replace without destroying the activity.

My feel is that one of the RxLoader implementations is the way to go instead of a homegrown solution here. `Loader` solves all of the lifecycle problems and is a widely tested solution, so we'd need **very** strong arguments ´for a solution not using it, which I don't see yet.

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -374,19 +351,67 @@ protected Intent navigateUp() {
...
+                .compose(RxTools.applySchedulers())
+                .subscribe(result -> {
+                    mIsWatching = (Boolean) result;
+                    supportInvalidateOptionsMenu();
+                }, e -> {});
+    }
+
+    public void isStarring() {
+        RepositoryService.isStarring(mApp, this, mRepoOwner, mRepoName)
+                .subscribe(result -> {
```
maniac103: No applySchedulers()?

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -284,69 +304,34 @@ public void onClick(View view) {
...
+            case R.id.tv_wiki_label:
+                intent = WikiListActivity.makeIntent(c, owner, name, null); break;
+            case R.id.tv_downloads_label:
+                intent = DownloadsActivity.makeIntent(c, owner, name); break;
+            case R.id.tv_releases_label:
+                intent = ReleaseListActivity.makeIntent(c, owner, name); break;
+            default:
+                if (view.getTag() instanceof Repository)
+                    intent = RepositoryActivity.makeIntent(c, (Repository) view.getTag());
+                break;
```
maniac103: This change is unrelated.

MattiaPrimavera:
Good point. Although, I checked on that one, and in my understanding it requires subclassing

app/src/main/java/com/gh4a/BaseActivity.java
```diff
@@ -440,7 +440,6 @@ public boolean onSearchRequested() {
 
     @Override
     public void onRefresh() {
-        supportInvalidateOptionsMenu();
```
maniac103: This is unrelated.

MattiaPrimavera:
P.s. binding to Fragments lyfecicle the same way I did for activities in previous commits, should be as much easy.

I specified I did not take care of that yet : 
> On fragments I still got to listen to their lyfecicle events ...

maniac103:
> Good point. Although, I checked on that one, and in my understanding it requires subclassing

There is not only one. https://github.com/evant/rxloader does require subclassing, https://github.com/kmdupr33/RxLoader or https://github.com/ArturVasilov/RxLoader don't. It'll probably require some experimentation to find out which one is the best one.

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
MattiaPrimavera: I moved apply schedulers into RxTools, since every call will define schedulers and error handler, I moved those to RxTools in order to compose with a single transformer that groups other transformers ...

maniac103: Yes, but just compare isStarring() to isWatching().

(BTW, please restrict method visibility, and a more appropriate name would be e.g. `updateWatchingStatus` and `updateStarringStatus` - `isWatching` implies the method returns a boolean that says whether the user is watching this repo)

MattiaPrimavera:
Since the cache is now linked to activities lifecycle, we d probably need RxLoader to be able emptying the cache as well ...

app/src/main/java/com/gh4a/BaseActivity.java
MattiaPrimavera: reverting ...

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
MattiaPrimavera: - right, I forgot removing that after moving it to **RxTools** 
- updating method signatures ...

app/src/main/AndroidManifest.xml
MattiaPrimavera: I'm sorry, you're right, I did not think of that ...

MattiaPrimavera:
> No, we shouldn't. Success is the normal case that doesn't need confirmation. 

Ok removing it. 

> What's the point of having a service for stuff that's only used once? I'd prefer non-reused code to stay at the place where it's used. 

Well, in my opinion you want those to appear within the activity class to inform the developer that it is used only there, although adding weight (lines of code) on the activity class file, and thus mixing concerns. 

Putting that in a separated file would **separate network concerns**, putting them out of the Activity file, making it easier for the developer to look for Async and Network operations, that would appear in the related Service class for every Activity / Fragment 

In order to indicate the use of those services is restrained to that Activity, I'd wrap both of them into a new package, and change the accessibility. 

What I mean is that when a new developer approaches the code (like me), he would find **lighter activity files**, **easier to read and follow**, and he would know that **UI** operations take probably place in that file whereas he would look into the **Service** file when inspecting async operations. 

In my opinion **separating concerns** in different files and using **packages** a little more would give a cleaner organisation, plus shorter class files means less time to look for information. When debugging the **UI instanciation** for instance, I personally do not want always to have the implementation code of other unrelated async operations in front of my eyes, this makes harder to directly point to what I need.

MattiaPrimavera:
P.S. I missed the **Progress** dialogs, could you pls give me an example on the app on how can i see one ?

maniac103:
> P.S. I missed the Progress dialogs, could you pls give me an example on the app on how can i see one ?

Given you've changed LabelListActivity, going to issues -> right drawer -> manage labels and editing a label seems like a good example ;-)

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
MattiaPrimavera: Agreed ... partially appreciated at least ? :P 

[Tunous]
> The change to switch statement is a good idea

maniac103: Yes, sure. I just like to keep apart unrelated changes. Such refactorings should go to separate pull requests, where they're more likely to be merged anyway.

MattiaPrimavera:
That's why I proposed showing a **success message** ... I did not see the **Progress bar** :S ... I thought the user had no feedback in success case ...

MattiaPrimavera:
I looked on **RxLoader** libraries, as asked, and choose to give a try at the [following](https://github.com/kmdupr33/RxLoader) 

```bash
compile 'com.philosophicalhacker.rxloader:rxloader:0.4.0'
```

which does not requiring **extending** activity classes.

Binding to lifecycle happens through the following line, within `RxTools`: 

```java
.compose(rxLoader.makeObservableTransformer(id));
``` 

Although the **Activities** / **Fragments** will not use that `Transformer` directly, but call `RxTools.handle()` function in order to both :

- attach error handler 
- bind to lifecycle 

```java
loadReadme()
  .compose(RxTools.handle(getActivity(), RepositoryService.LOAD_README, true))
  .subscribe(mReadmeConsumer, e -> {});
``` 

`true` value is passed to force reload

app/src/main/AndroidManifest.xml
Tunous: Yes, please revert unrelated changes. They just make it harder to review.

app/src/main/java/com/gh4a/Gh4Application.java
```diff
@@ -27,6 +27,7 @@
 
 import com.gh4a.fragment.SettingsFragment;
 import com.gh4a.utils.CrashReportingHelper;
+import com.gh4a.utils.rx.ObservableManager;
```
Tunous: Get rid of this line and the ObservableManager now that it's removed?

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -182,24 +131,42 @@ protected void onResultReady(Boolean result) {
...
-            getSupportLoaderManager().initLoader(LOADER_WATCHING, null, mWatchCallback);
-            getSupportLoaderManager().initLoader(LOADER_STARRING, null, mStarCallback);
+            this.isWatching();
+            this.isStarring();
         }
     }
 
+    public Observable<Repository> loadRepository() {
+        return RepositoryService.loadRepository(this, mRepoOwner, mRepoName)
+                .doOnNext(repository -> {
```
Tunous: I see no need to return Observable from this method. Instead, use subscribe method directly like in `isWatching`

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -182,24 +131,42 @@ protected void onResultReady(Boolean result) {
...
+        mApp = (Gh4Application) getApplicationContext();
 
         mActionBar = getSupportActionBar();
         mActionBar.setTitle(mRepoOwner + "/" + mRepoName);
         mActionBar.setDisplayHomeAsUpEnabled(true);
 
         setContentShown(false);
 
-        getSupportLoaderManager().initLoader(LOADER_REPO, null, mRepoCallback);
+        this.loadRepository().subscribe(o -> {}, e -> {});
```
Tunous: Code style: We do not prefix method with `this`. Applies to other places in code too.

app/src/main/java/com/gh4a/activities/issues_label/IssueLabelListActivity.java
```diff
@@ -244,21 +237,18 @@ public boolean onPrepareActionMode(ActionMode mode, Menu menu) {
         public boolean onActionItemClicked(ActionMode mode, MenuItem item) {
             switch (item.getItemId()) {
             case Menu.FIRST:
-                if (mLabel == mAddedLabel) {
-                    new AddIssueLabelTask(mLabel.editedName, mLabel.editedColor).schedule();
-                } else {
-                    new EditIssueLabelTask(mLabel.getName(), mLabel.editedName, mLabel.editedColor)
-                            .schedule();
-                }
+                if (mLabel == mAddedLabel) this.addIssue();
```
Tunous: Code style: Do not collapse if statements like this. Always use the full form with new lines and braces for both if and else branches. _Also remove `this.` prefix like mentioned earlier.

app/src/main/java/com/gh4a/activities/issues_label/IssueLabelListActivity.java
```diff
@@ -284,129 +274,30 @@ public void onDestroyActionMode(ActionMode mode) {
...
+            IssuesLabelService.addIssue(IssueLabelListActivity.this, getRootLayout(), mRepoOwner, mRepoName, mLabel.editedName, mLabel.editedColor)
+                .doOnError(error -> {
+                    mAdapter.remove(mAddedLabel);
+                    mAddedLabel = null;
+                })
+                .subscribe(result -> {
+                    forceLoaderReload(0);
+                    mAddedLabel = null;
+                    setResult(RESULT_OK);
+                }, e -> {});
```
Tunous: Use this error handler instead of `doOnError` above?

app/src/main/java/com/gh4a/activities/issues_label/IssueLabelListActivity.java
```diff
@@ -284,129 +274,30 @@ public void onDestroyActionMode(ActionMode mode) {
...
+        public void editIssueLabel() {
+            IssuesLabelService.editIssueLabel(IssueLabelListActivity.this, getRootLayout(), mRepoOwner, mRepoName, mLabel.getName(), mLabel.editedName, mLabel.editedColor)
+                .doOnError(error -> {
+                    mAdapter.remove(mAddedLabel);
+                    mAddedLabel = null;
+                })
+                .subscribe(result -> {
+                    forceLoaderReload(0);
+                    setResult(RESULT_OK);
+                }, e -> {});
```
Tunous: Same like before about `doOnError`.

app/src/main/java/com/gh4a/activities/issues_label/IssuesLabelService.java
```diff
@@ -0,0 +1,63 @@
...
+
+public class IssuesLabelService {
+    public static final int DELETE_ISSUE_LABEL = 0;
+    public static final int EDIT_ISSUE_LABEL = 1;
+    public static final int ADD_ISSUE_LABEL = 2;
+
+    public static Observable deleteIssueLabel(Activity activity, View rootView, String repoOwner, String repoName, String labelName) {
+        String errorMessage = activity.getString(R.string.issue_error_delete_label, labelName);
+
+        return Observable.fromCallable(() -> {
```
Tunous: Why Observable here? I feel like it would be better to simply use Completable since we don't return anything.

app/src/main/java/com/gh4a/activities/issues_label/IssuesLabelService.java
```diff
@@ -0,0 +1,63 @@
...
+            labelService.deleteLabel(repoOwner, repoName, URLEncoder.encode(labelName, "UTF-8"));
+            return true;
+        })
+        .compose(RxTools.onErrorSnackbar(activity, DELETE_ISSUE_LABEL, rootView, errorMessage, R.string.deleting_msg));
+    }
+
+    public static Observable editIssueLabel(Activity activity, View rootLayout, String repoOwner, String repoName, String oldLabelName, String newLabelName, String color) {
+        String errorMessage = activity.getString(R.string.issue_error_edit_label, oldLabelName);
+
+        return Observable.fromCallable(() -> {
```
Tunous: Completable?

app/src/main/java/com/gh4a/activities/issues_label/IssuesLabelService.java
```diff
@@ -0,0 +1,63 @@
...
+                    URLEncoder.encode(oldLabelName, "UTF-8"), label);
+            return true;
+        })
+        .compose(RxTools.onErrorSnackbar(activity, EDIT_ISSUE_LABEL, rootLayout, errorMessage, R.string.saving_msg));
+    }
+
+    public static Observable addIssue(Activity activity, View rootView, String repoOwner, String repoName, String labelName, String color) {
+        String errorMessage = activity.getString(R.string.issue_error_create_label, labelName);
+
+        return Observable.fromCallable(() -> {
```
Tunous: Completable?

app/src/main/java/com/gh4a/fragment/BaseFragment.java
```diff
@@ -0,0 +1,26 @@
...
+        super.onViewCreated(view, savedInstanceState);
+
+        mCompositeDisposable = new CompositeDisposable();
+    }
+
+    @Override
+    public void onDestroyView() {
+        super.onDestroyView();
+
+        if(mCompositeDisposable.isDisposed())
```
Tunous: Shouldn't this be inverted or I'm missing something? Like: `if (!mCompositeDisposable.isDisposed())`

Code style: Add space before `(` parenthesis.

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -137,7 +102,34 @@ public void onRefresh() {
...
+
+        this.reloadData();
+    }
+
+    public void reloadData() {
+        setContentShown(false);
+
+        Log.d("TEST", "reloadData callleeeed");
+        loadReadme()
+                .compose(RxTools.handle(getActivity(), RepositoryService.LOAD_README, true))
```
Tunous: RepositoryActivity didn't have these compose method calls. Could you explain why It's only used here?

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -148,8 +140,49 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+                readmeView.setTypeface(Typeface.DEFAULT, Typeface.ITALIC);
+            }
+            readmeView.setVisibility(View.VISIBLE);
+            progress.setVisibility(View.GONE);
+        };
+
+        // Fetch Readme file + PullRequests Count
+        loadReadme()
+                .compose(RxTools.handle(getActivity(), RepositoryService.LOAD_README))
+                .subscribe(mReadmeConsumer, e -> {
```
Tunous: Perhaps move these subscribe calls inside of `loadReadme`, `loadPullRequestCount` methods? Then there will be no need to create consumer fields.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+    public static final int UPDATE_STAR = 2;
+    public static final int UPDATE_WATCH = 3;
+    public static final int IS_STARRING = 4;
+    public static final int IS_WATCHING = 5;
+    public static final int LOAD_REPOSITORY = 6;
+
+    public static Observable loadReadme(Activity activity, String repoOwner, String repoName, String ref) {
+        Gh4Application app = (Gh4Application) activity.getApplicationContext();
+
+        return Observable.create(emitter -> {
```
Tunous: Use Single instead of Observable as there is just a single result value.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+                    emitter.onError(e);
+                }
+            } catch (IOException ioe) {
+                emitter.onError(ioe);
+            }
+        });
+    }
+
+    public static Observable loadPullRequestCount(Activity activity, Repository repository, String state) {
+        return Observable.create(emitter -> {
```
Tunous: Single.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+                emitter.onNext(count);
+                emitter.onComplete();
+            } catch(IOException ioe) {
+                emitter.onError(ioe);
+            }
+        });
+    }
+
+    public static Observable updateStar(Gh4Application app, Activity activity, String repoOwner, String repoName, boolean isStarring) {
+        return Observable.create(e -> {
```
Tunous: Single.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+            StarService starService = (StarService)
+                    Gh4Application.get().getService(Gh4Application.STAR_SERVICE);
+            RepositoryId repoId = new RepositoryId(repoOwner, repoName);
+            if (isStarring) starService.unstar(repoId);
```
Tunous: Code style.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+            else starService.star(repoId);
+
+            e.onNext(!isStarring);
+            e.onComplete();
+        })
+        .compose(RxTools.handle(activity, UPDATE_STAR));
+    }
+
+    public static Observable updateWatch(Activity activity, String repoOwner, String repoName, boolean isWatching) {
+        return Observable.create(e -> {
```
Tunous: Single.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            e.onNext(!isWatching);
+            e.onComplete();
+        })
+        .compose(RxTools.handle(activity, UPDATE_WATCH));
+    }
+
+    public static Observable isStarring(Activity activity, String repoOwner, String repoName) {
+        return Observable.create(e -> {
```
Tunous: SIngle.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+                e.onComplete();
+            } catch(Exception ex) {
+                e.onError(ex);
+            }
+        })
+        .compose(RxTools.handle(activity, IS_STARRING));
+    }
+
+    public static Observable isWatching(Activity activity, String repoOwner, String repoName) {
+        return Observable.create(e -> {
```
Tunous: Single.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,153 @@
...
+                e.onComplete();
+            } catch (Exception ex) {
+                e.onError(ex);
+            }
+        })
+        .compose(RxTools.handle(activity, IS_WATCHING));
+    }
+
+    public static Observable loadRepository(Activity activity, String repoOwner, String repoName) {
+        return Observable.create(e -> {
```
Tunous: Single.

app/src/main/java/com/gh4a/utils/rx/ObservableManager.java
```diff
@@ -0,0 +1,107 @@
+package com.gh4a.utils.rx;
```
Tunous: Delete this file?

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,99 @@
...
+                                .setAction(R.string.retry, (View.OnClickListener) activity)
+                                .show();
+                    });
+        };
+    }
+
+    public static <T> ObservableTransformer<T, T> handleError(Activity activity) {
+        return observable -> observable
+                .doOnError(error -> {
+                    if(activity instanceof BasePagerActivity) {
```
Tunous: Code style.

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
MattiaPrimavera: when no **cache refresh** or **loaders restart** is needed, in such cases as for -
- `updateStarring`
- `updateWatching`

services, the `compose` call can be hidden within the **Service** class (`RepositoryService`). 

Still I gotta update the code since : 
- `isStarring`
- `isWatching` 

services **do** need to be refreshed along with **repository** and **readme**. So there will be calls to `compose` within `RepositoryActivity` as well

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
MattiaPrimavera: That way I'd have to modify the functions `signature` in order to add a `boolean` parameter indicating wether to **refresh** the data, which would result in such call as `loadReadme(true)` or `loadReadme(false)` according to the need for **fresh / cached** data

app/src/main/java/com/gh4a/service/RepositoryService.java
MattiaPrimavera: Well noticed, It would be more appropriate ... but unfortunately the following [issue](https://github.com/kmdupr33/RxLoader/issues/24) on  `SingleTransformer` method exposure does not allow me transforming `Single` streams yet

app/src/main/java/com/gh4a/fragment/BaseFragment.java
MattiaPrimavera: Effectively ;) File removed

app/src/main/java/com/gh4a/service/RepositoryService.java
MattiaPrimavera: I think you meant 

```java
if (isStarring) {
  starService.unstar(repoId);
} else {
  starService.star(repoId);
}
```

I don't really see how would that make code clearer ... but it's done ;)

app/src/main/java/com/gh4a/activities/issues_label/IssuesLabelService.java
MattiaPrimavera: As for **Single**, we can move those instructions to **Single** and **Completable** but then convert them again to **Observable** in order to be able using `RxTools` Transformers, since `RxLoader` library has an **exposure** issue on `makeSingleTransformer(int, boolean)` method, and it does not provide at all methods to make `Transformers` on `Completable` objects (all this noticed after your remarks, but I guess still **preferable** than having to inherit from some Rx classes, like other **RxLoader** library implementations were proposing)

app/src/main/java/com/gh4a/activities/issues_label/IssueLabelListActivity.java
MattiaPrimavera: I'm sorry, I did not understand, could you detail that pls ?

Tunous: With this I meant the empty error handler:
`e -> {}`

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
Tunous: I would prefer that myself.

MattiaPrimavera: I like that too actually, gonna take care of it ;)

app/src/main/java/com/gh4a/service/RepositoryService.java
MattiaPrimavera: Do we stick with Observables everywhere while waiting for rxLoader lib update?

Tunous: I guess yes.

MattiaPrimavera: than what missing for closing this review comment?

Tunous: GitHub feature 😄. I don't see a way to close these manually. They only collapse if there was some code pushed that updated the mentioned line.

MattiaPrimavera: ^^` know I know, sorry :P

MattiaPrimavera:
Anything missing or we can go with the two **approaches** comparison (Loaders + AsyncTasks vs Rx) ?

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,102 @@
...
+            if(refresh) {
+                return observable
+                        .compose(handleError(activity))
+                        .compose(rxLoader.makeObservableTransformer(id, refresh));
+            } else return observable.compose(handle(activity, id));
+        };
+    }
+
+    // Error handler showing SnackBar
+    public static <T> ObservableTransformer<T, T> onErrorSnackbar(Activity activity, int id, View rootLayout, String errorMessage, int messageRes) {
```
maniac103: As above: pass in BaseActivity and add a Fragment variant. This allows calling activity.getRootLayout() from here without the need to pass it in.

Also the name (and comment) is off: it implies it (only) shows a Snackbar on error, but does much more. It's probably better to split the dialog and snackbar transformers.

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,102 @@
...
+        return observable -> {
+            final RxLoader rxLoader = new RxLoader(activity, ((AppCompatActivity)activity).getSupportLoaderManager());
+            return observable
+                    .compose(rxLoader.makeObservableTransformer(id))
+                    .doOnSubscribe(disposable -> showProgressDialog(activity, messageRes))
+                    .doOnTerminate(() -> dismissProgressDialog())
+                    .doOnError(error -> {
+                        Snackbar.make(rootLayout, errorMessage, Snackbar.LENGTH_LONG)
+                                .setAction(R.string.retry, (View.OnClickListener) activity)
```
maniac103: This still (I've commented on this before) is wrong: the click listener previously was ProgressDialogTask, and it re-scheduled the task. You now need to re-subscribe to the observable in that case.

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public static <T> ObservableTransformer<T, T> handleError(Activity activity) {
+        return observable -> observable
+                .doOnError(error -> {
+                    if (activity instanceof BasePagerActivity) {
+                        BasePagerActivity act = (BasePagerActivity) activity;
+                        act.setErrorViewVisibility(true);
+                    } else {
+                        // other cases
+                    }
```
maniac103: When passing in BaseActivity here, if-else can be dropped.

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,102 @@
...
+                });
+    }
+
+    public static <T> ObservableTransformer<T, T> applySchedulers() {
+        return observable -> ((Observable)observable).subscribeOn(Schedulers.io())
```
maniac103: Pretty sure the cast can be avoided by making the generic type `T extends Observable<?>`

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,102 @@
...
+                .observeOn(AndroidSchedulers.mainThread());
+    }
+
+
+    public static ProgressDialogFragment mProgressDialogFragment;
```
maniac103: As mentioned before: please don't keep static references to UI stuff. The fragment instance reference needs to be kept in the call chain of the transformer creation or in the transformer instance.

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    public static void dismissProgressDialog() {
+        if (mProgressDialogFragment.getActivity() != null) {
+            mProgressDialogFragment.dismissAllowingStateLoss();
+        }
+        mProgressDialogFragment = null;
+    }
+
+    public static void showProgressDialog(Activity activity, int messageRes) {
```
maniac103: This should accept BaseActivity, no need for casting in that case.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,156 @@
...
+    public static final int LOAD_README = 0;
+    public static final int LOAD_PULL_REQUESTS_COUNT = 1;
+    public static final int UPDATE_STAR = 2;
+    public static final int UPDATE_WATCH = 3;
+    public static final int IS_STARRING = 4;
+    public static final int IS_WATCHING = 5;
+    public static final int LOAD_REPOSITORY = 6;
+
+    public static Observable loadReadme(Activity activity, String repoOwner, String repoName, String ref, boolean refresh) {
+        Gh4Application app = (Gh4Application) activity.getApplicationContext();
```
maniac103: Gh4Application.get()

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,156 @@
+package com.gh4a.service;
+
+import android.app.Activity;
+import android.util.Log;
```
maniac103: Please remove import + debug statements.

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,156 @@
...
+            StarService starService = (StarService)
+                    Gh4Application.get().getService(Gh4Application.STAR_SERVICE);
+            try {
+                boolean isStarring = starService.isStarring(new RepositoryId(repoOwner, repoName));
+                e.onNext(isStarring);
+                e.onComplete();
+            } catch(Exception ex) {
+                e.onError(ex);
+            }
+        })
```
maniac103: The basic pattern seems pretty repetitive here:

```
public static Observable<...> doSomething() {
    return Observable.create(emitter -> {
        try {
            result = doSomething();
            emitter.onNext(result);
            emitter.onComplete();
        } catch (Exception e) {
            emitter.onError(e);
        }
    })
    .compose(RxTools.handle(activity, id, refresh));
}
```

This feels like there must be a more elegant way to write this, with a base method of some sort.

(Also note how I clarified variable names ;-) )

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -374,19 +346,57 @@ protected Intent navigateUp() {
...
 
+    public void isWatching(boolean refresh) {
+        RepositoryService.isWatching(this, mRepoOwner, mRepoName, refresh)
+                .subscribe(result -> {
+                    mIsWatching = (Boolean) result;
+                    supportInvalidateOptionsMenu();
+                }, e -> {});
+    }
+
+    public void isStarring(boolean refresh) {
```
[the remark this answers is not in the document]
Tunous: `loadWatchingStatus` vs `updateStarringStatus` should decide on one prefix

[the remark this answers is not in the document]
maniac103: Oops, right. `loadStarringStatus` it is then.

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -33,26 +31,20 @@
...
 import com.gh4a.utils.IntentUtils;
 import com.gh4a.utils.UiUtils;
-
 import org.eclipse.egit.github.core.Repository;
 import org.eclipse.egit.github.core.RepositoryBranch;
-import org.eclipse.egit.github.core.RepositoryId;
 import org.eclipse.egit.github.core.RepositoryTag;
-import org.eclipse.egit.github.core.service.StarService;
-import org.eclipse.egit.github.core.service.WatcherService;
-
```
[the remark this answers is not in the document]
MattiaPrimavera: This does not appear within `RepositoryActivity` anymore

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -182,24 +125,46 @@ protected void onResultReady(Boolean result) {
...
+            if (mInitialPage >= PAGE_REPO_OVERVIEW && mInitialPage <= PAGE_ACTIVITY) {
+                getPager().setCurrentItem(mInitialPage);
+                mInitialPage = -1;
+            }
+            setContentShown(true);
+            supportInvalidateOptionsMenu();
+        };
+
+        RepositoryService.loadRepository(this, mRepoOwner, mRepoName, refresh)
+                .subscribe(subscriber, e -> {});
```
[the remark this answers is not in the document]
MattiaPrimavera: If there is no matter error in the chain, it will be propagated to every next error handler in the chain until the error handler within the `subscribe` method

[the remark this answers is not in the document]
maniac103: Yes, but what does this have to do with removing an empty handler?

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -302,7 +267,14 @@ public void onRefresh() {
...
         setContentShown(false);
         invalidateTabs();
-        forceLoaderReload(0, 1, 2, 3);
+
+        loadRepository(true);
+
+        if (Gh4Application.get().isAuthorized()) {
+            isWatching(true);
+            isStarring(true);
+        }
```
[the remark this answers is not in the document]
MattiaPrimavera: No real difference with what was happening for `Loaders`: 3 lines to **restart** each one of them. 

I'll think of a more elegant solution too :)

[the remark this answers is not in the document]
maniac103: The previous solution was a single line one ;-)

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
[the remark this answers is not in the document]
MattiaPrimavera: sure? `load` prefix for **fetch** operations, whereas `update` prefix for **update** operations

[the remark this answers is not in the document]
maniac103: Please read all replies ;-)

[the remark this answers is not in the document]
MattiaPrimavera: ok I just saw the answer below

app/src/main/java/com/gh4a/service/IssuesLabelService.java
```diff
@@ -0,0 +1,63 @@
...
+import org.eclipse.egit.github.core.Label;
+import org.eclipse.egit.github.core.RepositoryId;
+import org.eclipse.egit.github.core.service.LabelService;
+import java.net.URLEncoder;
+import io.reactivex.Observable;
+
+public class IssuesLabelService {
+    public static final int DELETE_ISSUE_LABEL = 0;
+    public static final int EDIT_ISSUE_LABEL = 1;
+    public static final int ADD_ISSUE_LABEL = 2;
```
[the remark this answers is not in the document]
MattiaPrimavera: [Android Developers - Guide to App Architecture](https://developer.android.com/topic/libraries/architecture/guide.html#common_architectural_principles)
> The most important thing you should focus on is the separation of concerns in your app. It is a common mistake to write all your code in an Activity or a Fragment. Any code that does not handle a UI or operating system interaction should not be in these classes. Keeping them as lean as possible will allow you to avoid many lifecycle related problems. 

Just a question although, are Fragments shared among multiple activities in current architecture?

app/src/main/java/com/gh4a/service/RepositoryService.java
```diff
@@ -0,0 +1,156 @@
...
+public class RepositoryService {
+    public static final int LOAD_README = 0;
+    public static final int LOAD_PULL_REQUESTS_COUNT = 1;
+    public static final int UPDATE_STAR = 2;
+    public static final int UPDATE_WATCH = 3;
+    public static final int IS_STARRING = 4;
+    public static final int IS_WATCHING = 5;
+    public static final int LOAD_REPOSITORY = 6;
+
+    public static Observable loadReadme(Activity activity, String repoOwner, String repoName, String ref, boolean refresh) {
```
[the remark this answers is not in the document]
MattiaPrimavera: it breaks this line 

```java
        }).compose(RxTools.handle(activity, LOAD_README, refresh));
```

app/src/main/java/com/gh4a/service/IssuesLabelService.java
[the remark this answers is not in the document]
maniac103: > Just a question although, are Fragments shared among multiple activities in current architecture?

No, but fragments often have a shorter lifetime than their host activities.

Also your link doesn't seem applicable here. The loaders *are* a system API, thus operating system interaction.

app/src/main/java/com/gh4a/service/RepositoryService.java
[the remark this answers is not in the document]
maniac103: Then this line needs fixing. Giving up on strong typing is not an option.

app/src/main/java/com/gh4a/service/RepositoryService.java
MattiaPrimavera: Effectively ... what about this 

```java
    public static Observable loadRepository(Activity activity, String repoOwner, String repoName,
            boolean refresh) {
        return RxTools.runCallable(
            () -> ((org.eclipse.egit.github.core.service.RepositoryService)
                Gh4Application.get().getService(Gh4Application.REPO_SERVICE))
                .getRepository(repoOwner, repoName),
            activity,
            LOAD_REPOSITORY,
            refresh
        );
    }
```
+ `RxTools` method 

```java
    public static Observable runCallable(Callable call, Activity activity, int loaderId, boolean refresh) {
        return Observable.fromCallable(call)
                .compose(RxTools.handle(activity, loaderId, refresh));
    }
```

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -182,24 +125,46 @@ protected void onResultReady(Boolean result) {
...
-            getSupportLoaderManager().initLoader(LOADER_WATCHING, null, mWatchCallback);
-            getSupportLoaderManager().initLoader(LOADER_STARRING, null, mStarCallback);
+            isWatching(false);
+            isStarring(false);
         }
     }
 
+    public void loadRepository(boolean refresh) {
+        Consumer subscriber = repository -> {
+            mRepository = (Repository) repository;
```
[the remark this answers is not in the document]
MattiaPrimavera: solved with `Observable.create<String>()`

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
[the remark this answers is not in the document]
MattiaPrimavera: if I m not wrong this was resulting in an **ErrorHandlerNotImplemented** or something **Exception**, causing the app to crash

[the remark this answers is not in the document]
maniac103: Given there's [a variant of subscribe() without error handler parameter](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#subscribe-io.reactivex.functions.Consumer-) I have issues believing this is unavoidable ;-)

app/src/main/java/com/gh4a/service/IssuesLabelService.java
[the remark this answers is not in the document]
MattiaPrimavera: I was trying to make the point that including **150 lines** of code, all concerning **network calls**, would make Activity and Fragment code harder to read ... 

Furthermore if **fragments** are not shared among multiple **activities**, wouldn't it be easier to lean more on **packages** and use a **per-screen** structure, like creating **one package per screen**, that would contain both: 
- An Activity
- Its fragment(s)
- Its Service file 

**Screens packages** could than be regrouped by **theme**

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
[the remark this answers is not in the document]
MattiaPrimavera: That's what I thought too, but looking at the method code from RxJava library 

```
    /**
     * Runs the source observable to a terminal event, ignoring any values and rethrowing any exception.
     * <dl>
     *  <dt><b>Scheduler:</b></dt>
     *  <dd>{@code blockingSubscribe} does not operate by default on a particular {@link Scheduler}.</dd>
     * </dl>
     * @since 2.0
     */
    @SchedulerSupport(SchedulerSupport.NONE)
    public final void blockingSubscribe() {
        ObservableBlockingSubscribe.subscribe(this);
    }
```

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -137,7 +99,60 @@ public void onRefresh() {
...
+
+    public void loadReadme(boolean refresh) {
+        Consumer consumer = result -> { // Fill-in readme
+            Log.d("TEST-RE", "readme flatMap called");
+            setContentShown(true);
+            TextView readmeView = (TextView) mContentView.findViewById(R.id.readme);
+            View progress = mContentView.findViewById(R.id.pb_readme);
+            if (result != null)
+                mImageGetter.encode(getContext(), mRepository.getId(),
+                        String.valueOf(result));
```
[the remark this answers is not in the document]
MattiaPrimavera:
```
        ).doOnNext(result -> { // Fill-in readme
            setContentShown(true);
            TextView readmeView = (TextView) mContentView.findViewById(R.id.readme);
            View progress = mContentView.findViewById(R.id.pb_readme);
            if (result != null)
                mImageGetter.encode(getContext(), mRepository.getId(),
                        String.valueOf(result));

            if (result != null) {
                readmeView.setMovementMethod(UiUtils.CHECKING_LINK_METHOD);
                mImageGetter.bind(readmeView, String.valueOf(result), mRepository.getId());
            } else {
                readmeView.setText(R.string.repo_no_readme);
                readmeView.setTypeface(Typeface.DEFAULT, Typeface.ITALIC);
            }
            readmeView.setVisibility(View.VISIBLE);
            progress.setVisibility(View.GONE);
        }).subscribe(consumer -> {}, error -> {});
```

This seems working and should be executed on background ... 

I use `setContentShown` in `refresh` case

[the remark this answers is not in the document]
Tunous: It is **not** executed on a background thread. You can check that with `Looper.getMainLooper() == Looper.myLooper()`

app/src/main/java/com/gh4a/service/IssuesLabelService.java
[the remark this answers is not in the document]
Tunous: Perhaps, but that's not a topic for this pull request. @maniac103 has a point here. If an activity used multiple services, each with its own loader ids, then we would easily encounter loader id collision and unexpected behavior. For example `RepositoryService.LOAD_README` and `IssuesLabelService.DELETE_ISSUE_LABEL` both equal 0. I fear that this would break if both were used from same activity/fragment.

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,106 @@
...
+import android.view.View;
+import com.gh4a.BaseActivity;
+import com.gh4a.R;
+import com.gh4a.utils.UiUtils;
+import com.philosophicalhacker.lib.RxLoader;
+import java.util.concurrent.Callable;
+
+public class RxTools {
+    public static Observable runCallable(Callable call, BaseActivity activity, int loaderId, boolean refresh) {
+        return Observable.fromCallable(call)
```
Tunous: This shows `Unchecked assignment` warning here and in all the places that use this method. Adding generics should probably be enough to fix this. Like:

`public static <T> Observable<T> runCallbale(Callable<T> call, ...`

app/src/main/java/com/gh4a/utils/rx/RxTools.java
```diff
@@ -0,0 +1,106 @@
...
+                .compose(RxTools.handle(activity, loaderId, refresh));
+    }
+
+    public static Observable runCallable(Callable call, BaseActivity activity, int loaderId) {
+        return Observable.fromCallable(call)
```
Tunous: Same as above

app/src/main/java/com/gh4a/service/IssuesLabelService.java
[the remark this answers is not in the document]
maniac103: > For example RepositoryService.LOAD_README and IssuesLabelService.DELETE_ISSUE_LABEL both equal 0. I fear that this would break if both were used from same activity/fragment.

Yes, that's exactly what I meant.

> I was trying to make the point that including 150 lines of code, all concerning network calls, would make Activity and Fragment code harder to read ...

I disagree with that. IMHO overusing modularization can actually impair code readability. Putting half of the code of a method like loadReadme() into a separate class just doesn't seem useful to me.

Anyway, I think the ultimate goal should be converting the API bindings to use Rx anyway, at which point most of this discussion will be moot.

app/src/main/java/com/gh4a/service/IssuesLabelService.java
```diff
@@ -0,0 +1,68 @@
...
+                        URLEncoder.encode(oldLabelName, "UTF-8"),
+                        label
+                );
+        };
+
+        return RxTools.runOnErrorSnackBar(editIssue, activity, EDIT_ISSUE_LABEL,
+                errorMessage, R.string.saving_msg);
+    }
+
+    public static Observable addIssue(BaseActivity activity, String repoOwner, String repoName, String labelName, String color) {
```
Tunous: All methods in this class should specify result type for observables. For this method, it would be `Observable<Label>`. Also, make sure to fix any warning that would show after this change.

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -374,19 +327,75 @@ protected Intent navigateUp() {
         return UserActivity.makeIntent(this, mRepoOwner);
     }
 
+    private void loadWatchingStatus(boolean refresh) {
+        RepositoryService.loadWatchingStatus(this, mRepoOwner, mRepoName, refresh)
+                .subscribe(result -> {
+                    mIsWatching = (Boolean) result;
```
Tunous: Remove all no longer needed casts, give proper names instead of `result`. _Applies to other similar methods too._

app/src/main/java/com/gh4a/activities/RepositoryActivity.java
```diff
@@ -374,19 +327,75 @@ protected Intent navigateUp() {
...
+                        mIsWatching = !mIsWatching;
+                        supportInvalidateOptionsMenu();
+                    }, error -> {});
                 return true;
             case R.id.star:
                 MenuItemCompat.setActionView(item, R.layout.ab_loading);
                 MenuItemCompat.expandActionView(item);
-                new UpdateStarTask().schedule();
+
+                RepositoryService.setStarringStatus(this, mRepoOwner, mRepoName, mIsStarring)
```
Tunous: Unchecked call warning. Probably another problem with generics. You should go over all your changes and make sure to fix all these warnings.

maniac103:
Pulling out from the line comments:

>Anyway, I think the ultimate goal should be converting the API bindings to use Rx anyway, at which point most of this discussion will be moot.
> (NB: we could also consider using [this library](https://github.com/Meisolsson/GitHubSdk) for SDK bindings, but this seems like a huge task)

After having played with said library for a bit, I think I prefer converting to it **before** converting the loaders to Rx. We can make the loaders call blockingGet() in the initial conversion and remove them one by one later, instead of creating infrastructure for supporting the old way and throwing it away later. That binding library lacks a few features (e.g. reactions, or fetching paged responses wholesale), but in general looks pretty compatible.

MattiaPrimavera:
> I think I prefer converting to it before converting the loaders to Rx 

Why couldn't this be done in parallel ?

maniac103:
> Why couldn't this be done in parallel ?

Because we'd modify the same thing in parallel ... in other words, we'd create merge conflicts.

http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#subscribe-io.reactivex.functions.Consumer-

> Why would we throw Rx code away ?

Because we wouldn't have a need for the callable stuff anymore - it would be provided by the bindings already.

MattiaPrimavera:
As you guys like, but I find this a little pity, since I am available in contributing on this ... 

Moving to **Rx** now would be faster (since the **lib** your proposing is not ready, and needs to be updated, **time permitting** as you said... ) and would put the basis for the new **Rx** architecture, meaning once the library will be ready, it will be really easy to replace **Callable** with **Observables** provided by the library, whereas we could already take the advantages of **Rx** in the meanwhile, if you work on updating the library while I update the **front** to **Rx**, we'll replace the way data is loaded (replacing the **Observable.fromCallable()** call with a library call returning directly the **Observable**) 

So it's a little pity, considering we could probably handle this in parallel and share the efforts

maniac103:
Pushed my initial conversion results: https://github.com/slapperwan/gh4a/tree/new-bindings-prototype ... I'd appreciate some testing and listing of bugs you found. Bindings library is here: https://github.com/maniac103/githubsdk

Tunous:
@maniac103 I'll be running that branch on my main device to look for bugs. But from a quick look at the code, I can see that now there are many NullPointerException warnings now. Might want to do something about this.

Also perhaps create a separate pull request to declutter the conversation a bit.

maniac103:
> But from a quick look at the code, I can see that now there are many NullPointerException warnings now. Might want to do something about this.

Correct fix for that would be replacing Nullable with NonNull annotations for fields that are guaranteed to be present. Problem: it's not obvious from the documentation which ones are guaranteed to be present, and whether they're present or not likely also depends on the API endpoint returning the data.

> Also perhaps create a separate pull request to declutter the conversation a bit.

Actually I'd love to see a *new* pull request based on that branch, using the knowledge gained from this PR.

Tunous:
- [ ] My issues and pull requests always display the same data
- [ ] Crashes when opening issues or pull requests of a repository
- [ ] Repository start action loads infinitely

maniac103:
- [ ] User search shows null as follower and repo count (data isn't available in non-legacy search)

PR #700, opened 2017-09-08
When editing issue only update changed properties

### LINKED ISSUE #641: Editing only issue labels/milestones displays edited text in header
spectacle zu3326

The underlined text is what appears on issues even though I didn't modify issue's body or title. It doesn't appear when adding labels from the web interface.

Tunous:
Closes #641

### DIFF
--- a/github-api/src/main/java/org/eclipse/egit/github/core/service/IssueService.java
+++ b/github-api/src/main/java/org/eclipse/egit/github/core/service/IssueService.java
@@ -641,31 +641,41 @@ public class IssueService extends GitHubService {
 	protected Map<Object, Object> createIssueMap(Issue issue, boolean newIssue) {
 		Map<Object, Object> params = new HashMap<Object, Object>();
 		if (issue != null) {
-			params.put(FIELD_BODY, issue.getBody());
-			params.put(FIELD_TITLE, issue.getTitle());
+			String body = issue.getBody();
+			if (body != null) {
+				params.put(FIELD_BODY, body);
+			}
+
+			String title = issue.getTitle();
+			if (title != null) {
+				params.put(FIELD_TITLE, title);
+			}
+
 			List<User> assignees = issue.getAssignees();
 			if (assignees != null) {
 				List<String> assigneeLogins = new ArrayList<String>();
-				for (User assignee : assignees)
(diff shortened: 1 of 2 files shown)


github-api/src/main/java/org/eclipse/egit/github/core/service/IssueService.java
```diff
@@ -47,7 +30,24 @@
...
+import static org.eclipse.egit.github.core.client.IGitHubConstants.CHARSET_UTF8;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_COMMENTS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_EVENTS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_ISSUES;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_LEGACY;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_REACTIONS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_REPOS;
+import static org.eclipse.egit.github.core.client.IGitHubConstants.SEGMENT_SEARCH;
+import static org.eclipse.egit.github.core.client.PagedRequest.PAGE_FIRST;
+import static org.eclipse.egit.github.core.client.PagedRequest.PAGE_SIZE;
```
maniac103: Reordering seems unneeded (yes, I agree it's annoying, I just want to keep the diff to the original bindings as small as possible).

Tunous: Every single time before I commit I view whole diff and check if there is anything that I've missed, yet I still manage to always get these import changes unnoticed.

I probably have a checkbox to auto organize them enabled in a commit dialog... Will check and fix that later 😆

(I wish there was an option to use different styling rules per module in AS)

Tunous: Fixed and disabled the checkbox so I shouldn't make that mistake again.

maniac103:
Thanks for the fixup :)

PR #701, opened 2017-09-08
Readme: Update F-Droid button

dimqua:


Tunous:
F-droid has an official link as described [here](https://gitlab.com/fdroid/artwork/blob/master/badge/README.md), but it appears to be broken so I'm accepting this.

Thanks 👍

PR #707, opened 2017-09-08
Make README collapsible

### LINKED ISSUE #532: Option to collapse readme's
From https://github.com/slapperwan/gh4a/issues/233#issuecomment-68443773.

Tunous:
Closes #532


Tunous:
Preview

maniac103:
Maybe collapse by default and do lazy loading (that is, load on expand?)

Tunous:
Collapse by default could work, but lazy loading would make UX worse. It would increase the time you have to wait before you can start reading. And a lot of times what you want to do is to see the readme of a newly discovered repository.

maniac103:
Yes, but there's less data usage in case you're  *not* interested in the readme, which - in my usage patterns - is about 100% of all cases ;-)

Tunous:
Updated with lazy loading and collapsed state by default.

PR #715, opened 2017-09-10
Implement multiple tasks feature for issues

### LINKED ISSUE #710: Issue detail screen: enhancement
Wouldn't it be comfortable to have a button leading to **repository** home page, when looking at an issue? 

I often have the need to jump from issue to code ...

Tunous:
Closes #710

### DIFF
--- a/app/src/main/java/com/gh4a/activities/IssueActivity.java
+++ b/app/src/main/java/com/gh4a/activities/IssueActivity.java
@@ -16,18 +16,22 @@
 package com.gh4a.activities;
 
 import android.app.Activity;
+import android.app.ActivityManager;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
+import android.support.annotation.NonNull;
 import android.support.annotation.StringRes;
 import android.support.design.widget.AppBarLayout;
 import android.support.design.widget.CoordinatorLayout;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.content.Loader;
+import android.support.v4.util.ObjectsCompat;
 import android.support.v7.app.ActionBar;
 import android.support.v7.app.AlertDialog;
 import android.view.LayoutInflater;
@@ -49,6 +53,7 @@ import com.gh4a.loader.LoaderCallbacks;
 import com.gh4a.loader.LoaderResult;
 import com.gh4a.utils.ApiHelpers;
 import com.gh4a.utils.IntentUtils;
+import com.gh4a.utils.TaskUtils;
 import com.gh4a.utils.UiUtils;
 import com.gh4a.widget.BottomSheetCompatibleScrollingViewBehavior;
 import com.gh4a.widget.IssueStateTrackingFloatingActionButton;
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/gh4a/activities/IssueActivity.java
```diff
@@ -61,16 +66,36 @@
...
+                .putExtra(EXTRA_INITIAL_COMMENT, initialComment);
+    }
+
+    public static void startTask(Context context, String login, String repoName, int number,
+            IntentUtils.InitialCommentMarker initialComment) {
+        Intent intent = makeIntent(context, login, repoName, number, initialComment);
+        TaskUtils.startMatchingOrNewTask(context, intent, new TaskUtils.IntentMatcher() {
+            @Override
+            public boolean matches(@NonNull Bundle a, @NonNull Bundle b) {
+                b.setClassLoader(getClass().getClassLoader());
```
Tunous: This was needed to solve `NoClassDefFoundError` crashes when trying to read extras from task intent. It has something to do with `IntentUtils.InitialCommentMarker` but I found no other solutions.

You can replicate that crash by:
1. Comment out this line
2. Go to issue from news feed
3. Use recents to go back to news feed
4. Go to the same issue again

app/src/main/java/com/gh4a/activities/IssueActivity.java
```diff
@@ -113,17 +138,33 @@ protected void onResultReady(Boolean result) {
...
 
         ActionBar actionBar = getSupportActionBar();
-        actionBar.setTitle(getString(R.string.issue) + " #" + mIssueNumber);
+        String title = getString(R.string.issue) + " #" + mIssueNumber;
+        actionBar.setTitle(title);
         actionBar.setSubtitle(mRepoOwner + "/" + mRepoName);
         actionBar.setDisplayHomeAsUpEnabled(true);
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            // TODO: Better handle activity titles
```
Tunous: We might want to add a method for initializing this inside of BaseActivity after intent extras are read. It would then use getActivityTitle() to setup TaskDescription. But I'm not sure if this will be needed in all the other activities that won't be launched in separate tasks...

app/src/main/java/com/gh4a/activities/IssueListActivity.java
```diff
@@ -330,6 +330,7 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
                 Intent intent = IssueActivity.makeIntent(this, mRepoOwner, mRepoName,
                         issue.getNumber());
                 startActivityForResult(intent, REQUEST_ISSUE_AFTER_CREATE);
+                // TODO: Handle this
```
Tunous: I'd also like to have IssueEditActivity launched as a separate task when you are creating a new one (not when editing). But I'm not sure how to solve communication back to here.

app/src/main/java/com/gh4a/fragment/EventListFragment.java
```diff
@@ -368,15 +366,28 @@ public void onCreateContextMenu(ContextMenu menu, View v, ContextMenuInfo menuIn
 
         } else if (Event.TYPE_ISSUES.equals(eventType)) {
             IssuesPayload payload = (IssuesPayload) event.getPayload();
-            int issueNumber = payload.getIssue().getNumber();
+            final int issueNumber = payload.getIssue().getNumber();
             menu.add(getString(R.string.menu_issue, issueNumber))
-                    .setIntent(IssueActivity.makeIntent(getActivity(), repoOwner, repoName, issueNumber));
+                    .setOnMenuItemClickListener(new MenuItem.OnMenuItemClickListener() {
```
Tunous: This whole if block should get refactored. I also want to add some missing action and improve naming. But that's for another pull request so I think it's ok to leave these 2 new listeners like this for now.

app/src/main/java/com/gh4a/resolver/PullRequestDiffCommentLoadTask.java
```diff
@@ -83,6 +83,7 @@ protected Intent run() throws Exception {
                         false, mMarker);
             }
         } else {
+            // TODO: Handle this situation
```
Tunous: No ideas how to have this one launched inside of new task...

maniac103:
Superseded by #784

Issue #661, opened 2017-08-14
Images take long time to load


viktorstrate:
Some repositories, with many of images in their readme file, take up to a minute to load, with a decent internet speed. It is so slow, I somtimes open the repository in a browser, just because the images will load faster.

Tunous:
The problem is with https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/fragment/RepositoryFragment.java#L343 which calls https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/utils/HttpImageGetter.java#L393.

It's all executed in a single task and updates the text only after all images are loaded. This means that the more the images in a file the longer it will take to load. This is more problematic when viewing files in the file viewer as there the images will delay displaying of the whole file contents.

Example file I found that takes a longer while to open: https://github.com/google/flexbox-layout/blob/master/README.md

MattiaPrimavera:
any particular reason not to replace **AsyncTask** with **Picasso** library async image loading from a web url functionality on this one ?

maniac103:
Yes. After each image load, the HTML content of the readme file needs to be rendered to a format that can be displayed in a TextView again. Doing this after each image would be very inefficient.

Tunous:
Then something else must be slowing that file viewer because my example took much longer to load than other files. (Probably because it's big?)

For the TextView wouldn't it be possible to use layered drawables? (Initially set to layer with loading icon and after load switched to layer with image.) I saw examples of that on stack overflow before. That way it would be still rendered only once and we could use Picasso to do the actual loading of each image separately.

Sadly looking at the current code it'll require a fair bit of rewriting to get it working...

MattiaPrimavera:
@maniac103 if I correctly understood, that means **all the images are downloaded** before the placeholders gets replaced **all at once**, right ?

maniac103:
@MattiaPrimavera Right, exactly that.

PR #718, opened 2017-09-11
Fixing readme loading: Using webview

MattiaPrimavera:
I tried to use a **WebView** for solving #661 

It seems like it does work, no scrolling issues, since vertical one is disabled (still provided by the parent view) while the horizontal scroll stays enabled on the WebView. 

This leads to : 
- deleting **FillReadmeTask**
- deleting **lifecycle** methods linked to **HttpImageGetter** 
- delegation of html content rendering and images loading to the **WebView**


maniac103:
#661 already **is** fixed, though.

MattiaPrimavera:
What I meant is that in one of last discussion I've been told that WebView has been avoided to load the Readme and a custom solution had been prefer because of scrolling issues of nested views ... 

Wouldn't it be nice to delegate all this custom implemented behaviour to **WebView** ?

maniac103:
> What I meant is that in one of last discussion I've been told that WebView has been avoided to load the Readme and a custom solution had been prefer because of scrolling issues of nested views ...

Yes, among other things, in particular styling. For comparison, two screenshots of the Octodroid repo being displayed, with current solution and WebView:

Current:
screenshot_1505113971

WebView:
screenshot_1505114136

The WebView styling simply doesn't fit in as well, at least not without applying a lot of CSS magic.
Also, please note that we **can't** use horizontal scrolling, because horizontal swipes switch between tabs.

> Wouldn't it be nice to delegate all this custom implemented behaviour to WebView ?

Maybe, but we'll need to keep the custom handling for issue/PR comments anyway.

MattiaPrimavera:
> Maybe, but we'll need to keep the custom handling for issue/PR comments anyway.


Why is that? I obviously considered possibility of adjusting the solution, for the moment it was just trying to check if it was possible. 


Shouldn't the WebView handle horizontal scrolling for itself while swipe to move to other tabs should be provided by the activity?

maniac103:
> Why is that? I obviously considered possibility of adjusting the solution, for the moment it was just trying to check if it was possible.

Because having one WebView per issue comment view totally destroys performance. Yes, we've tried.

> Shouldn't the WebView handle horizontal scrolling for itself while swipe to move to other tabs should be provided by the activity?

Yes, but different horizontal swipe behaviour depending on where you swipe is bad from a UX perspective - it's just confusing.

MattiaPrimavera:
> Because having one WebView per issue comment view totally destroys performance. Yes, we've tried 

Yep I can imagine ... 

> Yes, but different horizontal swipe behaviour depending on where you swipe is bad from a UX perspective - it's just confusing. 

Isn't this what the app currently provides in **RepositoryFragment**?

maniac103:
> Isn't this what the app currently provides in RepositoryFragment?
No, RepositoryFragment doesn't scroll horizontally.

Tunous:
> Yes, but different horizontal swipe behaviour depending on where you swipe is bad from a UX perspective - it's just confusing.

Generally yes, but if the target area is easily distinguishable (like code blocks) then there should be no confusion. 

@MattiaPrimavera Did you test how horizontal scrolling works in code blocks with long text. From my experience WebView doesn't work well with ViewPager. (But maybe they fixed it)

Right now this pull request is a no go. Not before the WebView gets styling rules to look similar to what TextView based readme looks like. And I think we actually need these styling rules for the preview in comment editor so pull request improving that would be great on its own.

MattiaPrimavera:
@Tunous well I tried scrolling **horizontally** on a long text, if that's what you mean yes ... it looked like working 

> Right now this pull request is a no go. Not before the WebView gets styling rules to look similar to what TextView based readme looks like. And I think we actually need these styling rules for the preview in comment editor so pull request improving that would be great on its own. 

Totally agreed, the intention was not replacing with something uglier :P of course, but trying to see if a **WebView** could be used instead ;) 

> Because having one WebView per issue comment view totally destroys performance. Yes, we've tried 
@maniac103 Did you try as well within a `RecyclerView` ?

PR #723, opened 2017-09-11
Use code viewer text size setting for diff view in review.

maniac103:
@Tunous Thoughts? I'm unsure whether to do this or to just make the diff font a little smaller; it's currently a bit large for my taste.

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/timeline/DiffViewHolder.java
+++ b/app/src/main/java/com/gh4a/adapter/timeline/DiffViewHolder.java
@@ -2,6 +2,7 @@ package com.gh4a.adapter.timeline;
 
 import android.content.Context;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.graphics.Canvas;
 import android.graphics.Paint;
 import android.support.v4.content.ContextCompat;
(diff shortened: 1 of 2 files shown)


Tunous:
Both. Make the default diff font a little smaller (I think it's bigger than text in code viewer right now) and add that setting.

app/src/main/java/com/gh4a/adapter/timeline/DiffViewHolder.java
```diff
@@ -27,6 +30,9 @@
 
 class DiffViewHolder extends TimelineItemAdapter.TimelineItemViewHolder<TimelineItem.Diff>
         implements View.OnClickListener {
+    private static final float[] DIFF_SIZE_MULTIPLIERS = new float[] {
+            0.5F, 0.75F, 1F, 1.5F, 2F
```
Tunous: Maybe use values in `sp` instead? Will be easier to setup exact values. Right now smallest is almost unreadable while in code viewer it looks ok.

maniac103: As those are only multipliers over the (style-set) default text size, I think the size is set in sp already (and the scaling for sp is global anyway). If we want to keep that setting being used here, I think we'll need to adapt the default text size to match 'normal' code viewer size and adapt the multipliers to make the other sizes match.+

maniac103:
This should be better.

PR #725, opened 2017-09-11
Move watched repositories to Notifications screen

Tunous:
This reproduces the behavior present on GitHub web interface where notifications and watched repositories are grouped under 2 tabs in the same page.

Related to #578 and #624

### DIFF
--- a/app/src/main/java/com/gh4a/activities/home/HomeActivity.java
+++ b/app/src/main/java/com/gh4a/activities/home/HomeActivity.java
@@ -297,7 +297,7 @@ public class HomeActivity extends BaseFragmentPagerActivity implements
             case R.id.news_feed:
                 return new NewsFeedFactory(this, mUserLogin);
             case R.id.notifications:
-                return new NotificationListFactory(this);
+                return new NotificationListFactory(this, mUserLogin);
             case R.id.my_repos:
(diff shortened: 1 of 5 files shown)


maniac103:
I don't use notifications myself, so I'm not sure whether this is a good idea, but just to have mentioned it: I wonder whether having the watched repo list optional instead of a tab, that is, have it accessible via toolbar button only.

Tunous:
Personally, I would be happy with that option as I only use the notifications screen and rarely check my watched repositories. But, the aim of this change is to improve discoverability. Previously this entry was hidden in non-intuitive place. If we moved to overflow it would present a similar situation.

We could potentially move it to overflow menu for both Notifications and My Repositories screens but this seems more of a hack.

That's why I think tabs are the best option here. It's easy to find as it replicates tabbed interface in notifications screen on GitHub: https://github.com/watching

maniac103:
I totally forgot to answer you :-/ I wasn't thinking of overflow, but a real toolbar menu item, like the filter one in the repo list. I don't think that would cause a discoverability problem.
What I'm not a fan of (and would like to avoid) is putting more or less unrelated content in two tabs. I don't really like having two tabs already (as it somehow feels like a waste of screen space), but especially not in this case, even if Github does it :-/

PR #733, opened 2017-09-16
Improved about screens of profiles and repositories

Tunous:


Tunous:
Created pull request for this to have a place for easier overview and further discussions.

Tunous:
Suggestions:

1. Make click highlight span the whole row, only highlight icon if pressing on action icon.
2. Action icons should display icons for action to be performed. So for example, if you have watched the repository the icon should be a crossed-out eye.
3. The icon for following could be people with a + sign and unfollowing people with a - sign. That way followers can stay with people icon.
4. When long-pressing on action icon display toast telling what action will be performed?

Tunous:
This is how it works currently:
improved-about

maniac103:
> Make click highlight span the whole row, only highlight icon if pressing on action icon.

Yep, that's how it should behave.

> Action icons should display icons for action to be performed. So for example, if you have watched the repository the icon should be a crossed-out eye.

Thinking about it, I wonder whether we should adapt the icons at all or whether we should only indicate it by text, like I've done it for following a user. For watchers and stars, we could simply write '810 stars, including you'. 

> When long-pressing on action icon display toast telling what action will be performed?

That's a good idea.

Additionally, we should follow @smichel17 's suggesion and place a toolbar button for follow/unfollow in the follower list and a star/unstar button in the star list.

Tunous:
> Thinking about it, I wonder whether we should adapt the icons at all or whether we should only indicate it by text, like I've done it for following a user. For watchers and stars, we could simply write '810 stars, including you'.

Sounds good to me. We still need to find icons for following vs followers then.

maniac103:
Rebased to master

maniac103:
Implemented items 1 and 4.

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -385,15 +438,104 @@ private void loadPullRequestCount(boolean force) {
...
+        }
+
+        responseSingle.compose(RxUtils::doInBackground)
+                .subscribe(result -> {
+                    if (mIsWatching != null) {
+                        mIsWatching = !mIsWatching;
+                        mRepository = mRepository.toBuilder()
+                                .watchersCount(mRepository.watchersCount() + (mIsWatching ? 1 : -1))
+                                .build();
+                        updateStargazerUi();
```
Unpublished: Shouldn't this be updateWatcherUi?

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -385,15 +438,104 @@ private void loadPullRequestCount(boolean force) {
...
+                    }
+                }, error -> {
+                    handleActionFailure("Updating repo watching state failed", error);
+                    updateStargazerUi();
```
Unpublished: Same

Unpublished:
I don't know why watchersCount exists in api (as it is always the same as stargazers), but shouldn't we use subscribersCount? This seems to represent the real watchers count.

Tunous:
Yes, it might be outdated. I didn't sync the latest changes.

maniac103:
And/or I failed at rebasing. I'll rebase this as soon as I get some free time and look at this in the process.

maniac103:
Actually, watchersCount and stargazersCount will always be identical according to [1], so using subscribersCount indeed is the way to go.

[1] https://developer.github.com/v3/activity/watching/

app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
```diff
@@ -385,15 +438,104 @@ private void loadPullRequestCount(boolean force) {
...
+        }
+
+        responseSingle.compose(RxUtils::doInBackground)
+                .subscribe(result -> {
+                    if (mIsWatching != null) {
+                        mIsWatching = !mIsWatching;
+                        mRepository = mRepository.toBuilder()
+                                .subscribersCount(mRepository.subscribersCount() + (mIsWatching ? 1 : -1))
+                                .build();
+                        updateStargazerUi();
```
Unpublished: Just as a reminder this line and line 500 should still be `updateWatcherUi`.

maniac103: Done, thanks!

app/src/main/java/com/gh4a/adapter/RepositoryAdapter.java
```diff
@@ -59,7 +59,7 @@ public void onBindViewHolder(ViewHolder holder, Repository repository) {
         holder.tvLanguage.setText(repository.language() != null
                 ? repository.language() : mContext.getString(R.string.unknown));
         holder.tvForks.setText(String.valueOf(repository.forksCount()));
-        holder.tvStars.setText(String.valueOf(repository.watchersCount()));
+        holder.tvStars.setText(String.valueOf(repository.subscribersCount()));
```
Unpublished: Here we're actually looking for stars count...

maniac103:
Merging this, as it's been sitting far too long and seems to be fine for me after months of testing. If something is off, we can do follow-up fixes in master.

Issue #695, opened 2017-09-07
Move Timeline/Trending/Blog links to the news feed


smichel17:
I find it rather difficult to use scrollable navigation drawers. Octodroid's drawer is 3 entries longer than the height of my screen.

I think there's is a great opportunity to remove the `Public timeline`, `Trending Repos`, and `Github Blog`  links the drawer: Like the news feed, the are all about timely updates. So, they'd fit right in if they were moved tot he drop-down where you choose either your own or an organization's timeline to view (especially the public timeline, which is basically the news feed for all Github).

maniac103:
👍 for the global timeline idea, but Github blog and (especially) trending repos provide a completely different kind of information, so I don't think they fit well into the news feed.

Probably you just need a larger phone - the drawer fits fine on mine? ;-) Seriously though, scrollable drawers don't seem to be too uncommon (think e.g. Gmail or Google Maps), so I'm not sure avoiding it scrolling is a goal we should aim for at all costs.

Tunous:
I'm worried that moving public timeline here would make it hard to discover. I myself wouldn't think about searching for it here. But I rarely used that option.

smichel17:
I agree they are common. Common and terrible.

I guess I should also clarify that I'm talking in particular about scrollable *navigation* drawers. Having UI elements in a consistent position means I can build muscle memory and navigate faster as I use the app more. A scrollable drawer throws this all away.

In the case of a scrollable drawer for user generated content, the content will change anyway so you're not going to be able to build that muscle memory the same way (ex: Simpletask has a drawer for filtering tasks based on lists and tags that I never minded).

Issue #684, opened 2017-09-05
Can't load the corresponding content for filter/order setting


WanderMax:
I choose the start page : My repositories, and set filter "Starred" and sort order "Most recently pushed".
However, when I open the app, the loaded content is always "My repositories" with one strange order.
In previous version, app can load the "Starred" content with "Most recently pushed" order.
Thanks.

Tunous:
Same issue as #746 ?

maniac103:
It's extremely likely, yes.

PR #734, opened 2017-09-16
Move starred repositories to bookmarks screen

Tunous:
stars-and-bookmarks

Related to #578, #624 and #725

### DIFF
--- a/app/src/main/java/com/gh4a/activities/RepositoryListActivity.java
+++ b/app/src/main/java/com/gh4a/activities/RepositoryListActivity.java
@@ -105,8 +105,8 @@ public class RepositoryListActivity extends FragmentContainerActivity implements
     protected void onPrepareRightNavigationDrawerMenu(Menu menu) {
         if (mFragment != null) {
             mFilterDrawerHelper.selectFilterType(menu, mFragment.getFilterType());
-            mSortDrawerHelper.selectSortType(menu,
-                    mFragment.getSortOrder(), mFragment.getSortDirection());
+            mSortDrawerHelper.selectSortType(menu, mFragment.getSortOrder(),
+                    mFragment.getSortDirection(), false);
         }
     }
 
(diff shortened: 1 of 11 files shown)


Tunous:
Myself I would prefer to have this as a separate entry in the drawer as seen in #578. But decided to create this pull request to hear your opinions.

smichel17:
I would prefer this implementation because, as I said in #695, I think scrollable navigation drawers are terrible ux:

> Having UI elements in a consistent position means I can build muscle memory and navigate faster as I use the app more. A scrollable drawer throws this all away.

maniac103:
I don't necessarily agree to the scrolling drawer argument (we have _a lot_ of stuff to show, so we probably won't be able to avoid having a scrolling drawer - it's already scrolling in @Tunous example), but I think this implementation is fine, given that having a 'My repositories' + a 'Starred repositories' drawer item is somewhat inconsistent and potentially confusing as well.

Tunous:
So, let's go for #725 and this + close #578 and #624?

smichel17:
> we have a lot of stuff to show, so we probably won't be able to avoid having a scrolling drawer - it's already scrolling in @Tunous example

Also in @Tunous's example, if `Public timeline`, `Trending repos`, and `Github blog` were collapsed into a single entry, it would not scroll. My phone is slightly smaller than that, so we would have to remove those 3 entries (#695), not collapse them, but it's not that far off. However, this is now off-topic for this PR so I'll shut up about it now :)

maniac103:
I just noticed via #684 that this PR in its current state loses sort order options.

Tunous:
Is it OK to add it as submenu similar to the one for filtering notifications? I think that will fit better than right-side menu.

maniac103:
Yeah, it surely is.

Tunous:
Added.

Tunous:
@maniac103 perhaps it's time to decide on what to do with this. I think the option I mentioned [here](https://github.com/slapperwan/gh4a/pull/734#issuecomment-332198696) is the way to go.

maniac103:
This one is fine with me; I have more issues with #725.

PR #736, opened 2017-09-20
Notifications

### LINKED ISSUE #42: Notifications
Another feature that I'd love to see is notifications.

obviously you would get spammed if you made it so that it notifies you every time someone push and stuff.

A great way to solve this would be able to 'mark' a report for notifications so that you only recieve notifications from those repos!

Tunous:
Implemented using https://github.com/evernote/android-job for backward compatibility without the need to have Google Play Services.

- Each GitHub notifications gets its own notification and all of them are bundled together (Not sure how it works on older API, need to test that)
- There is a setting to toggle them, which I have set to be disabled by default
- There is a setting to change synchronization interval ranging from 15min (lowest possible value) to 6 hours (can add higher values)
- Clicking on notification opens its corresponding screen and automatically dismisses it from notification drawer
- All notifications belong to a single channel on Android O so users can customize here a priority, sound, vibration, etc.

Potential improvements:
- Auto-mark as read when clicking on notifications (could be controlled by setting)
- Button to manually mark as read (unsubscribe?)
- Include launcher badges (I plan to create a separate pull request for that once this one is merged)
- Better icon. I've used the launcher icon but it would be nice if someone could create a vector drawable with a similar shape. @slapperwan perhaps you still have contact with the guy that made the launcher icon (Would be also nice to create adaptive icon)
- Display icon of repository owner on each notification. (@maniac103 I'm not sure myself how to make it work well. We need something to load a Bitmap and the loading must finish before Job finishes.)
- Support notifications for multiple accounts
- Handle logged-out case (do not check for notifications when logged out, stop job when logging out)

Closes #42


app/src/main/java/com/gh4a/Gh4Application.java
```diff
@@ -158,6 +166,31 @@ public void onCreate() {
         mServices.put(STAR_SERVICE, new StarService(mClient));
         mServices.put(USER_SERVICE, new UserService(mClient));
         mServices.put(WATCHER_SERVICE, new WatcherService(mClient));
+
+        createNotificationChannels();
```
maniac103: I think it would be better to do this asynchronously to not delay startup for this. I assume you do this instantly to provide the respective controls in the settings app before the first notification arrives? Maybe it also would make sense to do it when enabling notifications in our settings?

app/src/main/java/com/gh4a/fragment/SettingsFragment.java
```diff
@@ -178,7 +205,8 @@ public OpenSourceComponentListDialog(Context context) {
             { "HoloColorPicker", "https://github.com/LarsWerkman/HoloColorPicker" },
             { "Material Design Icons", "https://github.com/google/material-design-icons" },
             { "PrettyTime", "https://github.com/ocpsoft/prettytime" },
-            { "SmoothProgressBar", "https://github.com/castorflex/SmoothProgressBar" }
+            { "SmoothProgressBar", "https://github.com/castorflex/SmoothProgressBar" },
+            { "Android-Job", "https://github.com/evernote/android-job" }
```
maniac103: This also needs to be added to README.md.

app/src/main/java/com/gh4a/job/NotificationsJob.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @NonNull
+    @Override
+    protected Result onRunJob(Params params) {
+        try {
+            List<Notification> notifications =
+                    NotificationListLoader.loadNotifications(false, false);
+
+            if (notifications.isEmpty()) {
+                return Result.SUCCESS;
```
maniac103: You'd need to cancel existing notifications before returning here, wouldn't you?

app/src/main/java/com/gh4a/job/NotificationsJob.java
```diff
@@ -0,0 +1,129 @@
...
+            }
+
+            NotificationManagerCompat notificationManager =
+                    NotificationManagerCompat.from(getContext());
+            notificationManager.cancelAll();
+
+            int accentColor = UiUtils.resolveColor(getContext(), R.attr.colorAccent);
+
+            showSummaryNotification(notificationManager, notifications.size(), accentColor);
+            for (Notification notification : notifications) {
```
maniac103: I think it would be better to show one sub-notification per repo instead of per notification if the total number of notifications exceeds a certain threshold (e.g. 10).

app/src/main/java/com/gh4a/job/NotificationsJob.java
```diff
@@ -0,0 +1,129 @@
...
+            PendingIntent pendingIntent =
+                    PendingIntent.getActivity(getContext(), 0, intent, 0);
+            builder.setContentIntent(pendingIntent);
+        }
+
+        int id = ID_SUMMARY + 1;
+        try {
+            id = Integer.parseInt(notification.getId());
+        } catch (NumberFormatException e) {
+            // ignored
```
maniac103: If this fails (it probably won't, though) we'll nuke another notification. Suggestion: Drop the ID parsing and use the offset into the notification array (+ 1 for summary) as notification ID.

app/src/main/java/com/gh4a/job/NotificationsJob.java
```diff
@@ -0,0 +1,129 @@
...
+        notificationManager.notify(ID_SUMMARY, new NotificationCompat.Builder(
+                getContext(), Gh4Application.CHANNEL_GITHUB_NOTIFICATIONS)
+                .setSmallIcon(R.drawable.octodroid)
+                .setGroup(GROUP_ID_GITHUB)
+                .setGroupSummary(true)
+                .setColor(accentColor)
+                .setContentTitle(title)
+                .setContentText(text)
+                .setNumber(numNotifications)
+                .build());
```
maniac103: Probably should add an InboxStyle for pre-L notifications here.

app/src/main/res/values/strings.xml
```diff
@@ -518,4 +518,13 @@
     <string name="wiki_print_document_title">Wiki page %1$s in %2$s/%3$s</string>
 
     <string name="repository_owner_avatar">Repository owner avatar</string>
+    <string name="channel_notifications_name">GitHub notifications</string>
+    <string name="channel_notifications_description">Notifications received from GitHub.</string>
```
maniac103: I haven't seen the style used for other notification channel descriptions, but my gut feeling is that the trailing period should be removed.

maniac103:
First of all: Great, thanks!

Some comments to the questions raised:

> Each GitHub notifications gets its own notification and all of them are bundled together (Not sure how it works on older API, need to test that)

As mentioned in the code comments, I think it would be better to collapse the notifications by-repo if the total number exceeds a certain threshold to not overwhelm users with lots of notifications.

> There is a setting to toggle them, which I have set to be disabled by default
> There is a setting to change synchronization interval ranging from 15min (lowest possible value) to 6 hours (can add higher values)

👍 I'm not sure higher values are needed, maybe add 12 and 24 hours?

> Auto-mark as read when clicking on notifications (could be controlled by setting)

When adding this, it certainly needs a setting; it feels a little heavy handed to do it automatically. But with the below, I'm not even sure it's needed.

> Button to manually mark as read (unsubscribe?)

That (I assume by button you mean notification action?), + mark all as read for the summary notification.

> Include launcher badges (I plan to create a separate pull request for that once this one is merged)

👍 

>Better icon. I've used the launcher icon but it would be nice if someone could create a vector drawable with a similar shape. @slapperwan perhaps you still have contact with the guy that made the launcher icon (Would be also nice to create adaptive icon)

I made [an SVG](https://github.com/slapperwan/gh4a/blob/master/artwork/octodroid.svg) from the icon already. It should be relatively easy to convert this to black and scale it down, just need to manually remove the details for the latter.

> Display icon of repository owner on each notification. (@maniac103 I'm not sure myself how to make it work well. We need something to load a Bitmap and the loading must finish before Job finishes.)

Given the job runs on a background thread, we'd probably just need some plumbing to AvatarHandler to return the Bitmap for a given User right away. I can do that.

> Support notifications for multiple accounts

Yeah, that's needed.

> Handle logged-out case (do not check for notifications when logged out, stop job when logging out)

The latter implies the former though (as long as multi-account handling isn't implemented), doesn't it?

app/src/main/java/com/gh4a/Gh4Application.java
Tunous: > Maybe it also would make sense to do it when enabling notifications in our settings?

That's a good idea.

app/src/main/java/com/gh4a/job/NotificationsJob.java
Tunous: Perhaps, or perhaps not. I don't want the user to lose notifications. Though, if they marked them as read from other place perhaps it's ok to clear them... I'm not really sure.

maniac103: I'd compare this to an email (IMAP) client which I'd also expect to clear email notifications when I've read them elsewhere. At least that's what I made the CM/LineageOS email client do. ;-)

Tunous:
> When adding this, it certainly needs a setting; it feels a little heavy handed to do it automatically. But with the below, I'm not even sure it's needed.

Auto marking as read would be useful even with button because clicking on a notification dismisses it.

> That (I assume by button you mean notification action?), + mark all as read for the summary notification.

Yes that's what I mean. 

> The latter implies the former though (as long as multi-account handling isn't implemented), doesn't it?

No, what I meant is to 

- add a check in Application class to cancel instead of rescheduling if user is logged out
- Cancel the job when user performs log out action, otherwise the job could try to run before the check in Application disables it

maniac103:
TODOs left from my comments:
- [ ] InboxStyle for summary
- [ ] Collapse notifications per-repo if there are a lot of them (needs thinking about click intent for those)

Tunous:
> Create a public version of the summary notification (or make it public right away)

What does that mean?

maniac103:
>> Create a public version of the summary notification (or make it public right away)

> What does that mean?

The default notification visibility is [private](https://developer.android.com/reference/android/app/Notification.html#VISIBILITY_PRIVATE), which means that for our notifications the context text line isn't shown on the secure lock screen. We'd either need [a public version](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.Builder.html#setPublicVersion(android.app.Notification) to be shown on the secure lock screen, or, as the number of unread notifications isn't exactly private, just make it [public](https://developer.android.com/reference/android/app/Notification.html#VISIBILITY_PUBLIC).

Tunous:
Added:
- Handling of logged out case
- Button to mark notification as read
- Option to auto-mark notifications as read when opening them

I'm not sure if it's a good idea to put these in BrowseFilter but at least they work correctly 😃 

I've also checked all apps on Android O emulator to see how they write descriptions (with a comma or not) and surprisingly none of them had any descriptions... I would leave a comma here myself because this place looks like it could be intended for detailed multiline text. It's a description after all.

Tunous:
> Collapse notifications per-repo if there are a lot of them (needs thinking about click intent for those)

Make this redirect to notification screen and scroll to the corresponding repository (with highlight like comments)?

maniac103:
> Make this redirect to notification screen and scroll to the corresponding repository (with highlight like comments)?

:+1: Makes sense.

Tunous:
Some more suggestions:

- Make it so the notification plays sound/vibration
- Try to extract comment/issue body if possible and display this as a text

Tunous:
- rebased
- added notification defaults so there is sound and vibration
- added inbox style for summary notification
- fixed some bugs

Now that the summary notification has inbox style it shouldn't be made public.

And there is probably no need to add per-repo notifications. Older versions show only summary notification (untested yet but I assume that's how it should work) and bundled notifications only show some of the notifications in the expanded state. Dismissing them one by one shows hidden ones. I think this works well.

Tunous:
Bug: Summary notification reappears on next job run even if previously displayed notification has been marked as read.

maniac103:
> added notification defaults so there is sound and vibration

Should there be an option for that?

> Now that the summary notification has inbox style it shouldn't be made public.

Right, but (IMHO) it should get a public version that only says 'xy new notifications'.

> And there is probably no need to add per-repo notifications. Older versions show only summary notification (untested yet but I assume that's how it should work) and bundled notifications only show some of the notifications in the expanded state. Dismissing them one by one shows hidden ones. I think this works well.

Hmm, that's probably right, looking at e.g. how the email app behaves. I'm wondering though whether we shouldn't only notify about notifications that the user hasn't seen (that is, store time of last visit of notification page and only notify of notifications newer than that). Not sure.

Tunous:
> I'm wondering though whether we shouldn't only notify about notifications that the user hasn't seen (that is, store time of last visit of notification page and only notify of notifications newer than that). Not sure.

We definetly should. Right now it replays sound on every job run, even if notifications don't change. This is really annoying. 

> Right, but (IMHO) it should get a public version that only says 'xy new notifications'.

Sure.

maniac103:
> We definetly should. Right now it replays sound on every job run, even if notifications don't change. This is really annoying.

We could also opt to not kill the summary notification unconditionally, but only if there are no notifications, and then set [this flag](https://developer.android.com/reference/android/app/Notification.html#FLAG_ONLY_ALERT_ONCE) if there's no notification with timestamp newer than last check time.

Tunous:
> > added notification defaults so there is sound and vibration
> 
> Should there be an option for that?

I'm not sure. On Android O, everything is configurable through channel. Android Nougat also has setting to show notifications silently or disable LED. But probably it's worth doing it for the older versions. Have to decide which options to make configurable (just sound or sound and vibration or LED too?)

Tunous:
We could also colorize the notifications with color matching icon for better looks.

smichel17:
I don't think github notifications are important enough to warrant commoditization.

Tunous:
I've improved the alert behavior. Now it should behave correctly and not annoy users with re-appearing notifications.

@smichel17 I meant coloring for some text and small icon, not the background. Here is how it looks now:
notifications

I like this small icon with details, looks very nice. But still, need a new drawable for it with smaller size and non-transparent color. (On my device it's not tinted when located in status bar) @maniac103 is it ok to just copy octodroid_bg and modify these values or some other changes are needed?

smichel17:
ah, yeah, that makes sense.

maniac103:
@Tunous I don't understand the question? The image should be - IIRC - 16x16dp and black on transparent.

The last check + local only stuff feels a bit weird to me ATM, because I'd base the decision on last read timestamp, not last check timestamp, I think (would also need to add a delete intent to update that read timestamp). I'll think about this further.

Tunous:
> I don't understand the question? The image should be - IIRC - 16x16dp and black on transparent.

Nvm. I got it.

maniac103:
Thought about this some more and what I think makes most sense is this:
- Have a 'last read timestamp' that's updated when either visiting the notification page or when swiping away the summary notification (-> delete intent)
- Only show a new notification if there's a GH notification that are newer than that timestamp (idea being that if you're clearing a notification, you implicitly acknowledge its content and don't want to be notified of it again). Don't clear existing notification if there isn't a newer GH notification though.
- If there's a GH notification newer than that, still notify of all pending GH notifications, not only those newer than timestamp (similar to email apps: have notification for 3 emails, swipe it away, next time you'll get a notification for 4 emails, not for a single one.
- Don't fiddle with setLocalOnly(). I don't want the notification to be cleared from my watch just because I didn't read it until the next job run.
- Go back to using GH notification IDs as Android notification IDs. This allows avoiding to re-post them by calling [getActiveNotifications](https://developer.android.com/reference/android/app/NotificationManager.html#getActiveNotifications()) and matching existing Android notifications with the new notification list. If everything is equal, there's no need to re-post the summary notification; if there's a change, we can now determine whether the summary needs ALERT_ONLY_ONCE (all existing notification timestamps are <= saved timestamp, thus there's nothing new) or not.

Does this make sense? If yes, I can start implementation of this.

Tunous:
> Have a 'last read timestamp' that's updated when either visiting the notification page or when swiping away the summary notification (-> delete intent)

What if you dismiss all notifications one by one instead? Will it still fire that intent after last one is dismissed?

In addition to what you said I think it would be good to not clear notifications at all and conditionally add ALERT_ONLY_ONCE to each notification.

maniac103:
> What if you dismiss all notifications one by one instead? Will it still fire that intent after last one is dismissed?

Probably not, we'd have to code this manually (delete intent on each single notification). I'd like to avoid the complexity of having to remember read status for each notification locally though (that's what email apps are doing) and instead would add delete intent handling just so that swiped away notifications are marked as read if that option is enabled and accept them to reappear otherwise. Maybe it also would make sense to unconditionally mark them as read in that case? Not sure.

> In addition to what you said I think it would be good to not clear notifications at all and conditionally add ALERT_ONLY_ONCE to each notification.

Do you really want to play a sound for *each* of the single notifications? Not sure what NotificationManager does in that case.
In either case, we'll need to clear notifications that no longer are present in the list returned by Github.

BTW, I think the BrowseFilter addition really is a bit hacky. We should use an IntentService to handle that stuff instead. Services run from PendingIntents are allowed to run for a few minutes after being invoked on O, so we shouldn't have a problem there.

Tunous:
> Do you really want to play a sound for each of the single notifications? Not sure what NotificationManager does in that case.

No, that's why I added `.setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_SUMMARY)`. Adding ALERT_ONLY_ONCE would just prevent from waking my screen, showing desktop notification with the same notifications again (I hope)

> BTW, I think the BrowseFilter addition really is a bit hacky. We should use an IntentService to handle that stuff instead. Services run from PendingIntents are allowed to run for a few minutes after being invoked on O, so we shouldn't have a problem there.

Agreed, I just had no better ideas.

Tunous:
I think that clicking on mark read action should also dismiss the notification.

Tunous:
@maniac103 Pushed some changes on top of yours. I'm not really sure what else could/should be done here so I'm leaving it up to you.

Right now I'm not sure about the behavior of "Auto mark read" setting. Before it was ok because it was only for a single notification. But now that it's for the whole repository it seems to be a bit too dangerous option. Maybe it should take effect only if there is a single notification for a repository?

Tunous:
1. When marking last notification as read with button it disappears but summary notification shows instead
2. The text in repository notifications is a bit misleading. When they are collapsed we can see only a line for a single notification and no indicator that there are more. This is misleading because someone could assume that there is only one notification for that repository and clicking on it would redirect to the corresponding link instead of the list of notifications.

maniac103:
Rebased to master and added some more fixes, in particular the mentioned 'seen' handling (see https://github.com/slapperwan/gh4a/pull/736/commits/f75868a07818f9b14ff7bf4df5f6fad1767639cc)

> Right now I'm not sure about the behavior of "Auto mark read" setting. Before it was ok because it was only for a single notification. But now that it's for the whole repository it seems to be a bit too dangerous option. Maybe it should take effect only if there is a single notification for a repository?

Or we nuke the option and the special single notification handling altogether and always send the user to the notification page?


> When marking last notification as read with button it disappears but summary notification shows instead

What do you mean here?

> The text in repository notifications is a bit misleading. When they are collapsed we can see only a line for a single notification and no indicator that there are more. This is misleading because someone could assume that there is only one notification for that repository and clicking on it would redirect to the corresponding link instead of the list of notifications.

That's the way MessagingStyle works, though, and I don't see any good alternative. That's another point for the suggestion above, as 'send user to the list of messages' is what messaging apps do in that case as well.

Tunous:
> Or we nuke the option and the special single notification handling altogether and always send the user to the notification page?

Yeah, that should work.

>> When marking last notification as read with button it disappears but summary notification shows instead
>
>What do you mean here?

1. Single notification appears
2. I click on the mark as read button which also automatically dismisses that notification
3. Summary notification appears saying that there is one unread notification (but it doesn't list it)

Tunous:
I just noticed: When user presses on mark as read button from notifications we should pass a timestamp. Otherwise they could accidently mark unseen notifications. I think API supports this.

app/src/main/java/com/gh4a/fragment/NotificationListFragment.java
```diff
@@ -255,6 +263,33 @@ private void markAsRead(Repository repository, Notification notification) {
...
+            return;
+        }
+
+        for (int i = 0; i < notifications.size(); i++) {
+            NotificationHolder holder = notifications.get(i);
+            if (holder.notification == null) {
+                Repository repo = holder.repository;
+                if (repoOwner.equals(repo.getOwner().getLogin())
+                        && repoName.equals(repo.getName())) {
+                    scrollToAndHighlightPosition(i);
```
Tunous: Missing break statement

app/src/main/java/com/gh4a/resolver/BrowseFilter.java
```diff
@@ -11,11 +11,13 @@
...
 
 public class BrowseFilter extends AppCompatActivity {
+    private static final String EXTRA_INITIAL_COMMENT = "initial_comment";
+
     public static Intent makeRedirectionIntent(Context context, Uri uri,
             IntentUtils.InitialCommentMarker initialComment) {
         Intent intent = new Intent(context, BrowseFilter.class);
         intent.setData(uri);
-        intent.putExtra("initial_comment", initialComment);
+        intent.putExtra(EXTRA_INITIAL_COMMENT, initialComment);
```
Tunous: We might want to share this constant. It feels a bit weird to have it used only in a single place.

Tunous:
Didn't find any more issues (other than these mentioned) from review and testing. Should be ready for merging.

app/src/main/java/com/gh4a/resolver/BrowseFilter.java
maniac103: I've added a commit that moves the intent extra access to BrowseFilter and shares the constant.

maniac103:
Fixed the two mentioned things and didn't notice any issues myself either.

PR #741, opened 2017-09-23
Improved milestone selection dialog

### LINKED ISSUE #570: Assign closed milestone to issue/pr
On GitHub closed milestone can be assigned by selecting closed tab in the pop-up.

Here we could either list all closed milestones under open or modify the milestone selection dialog to contain open and closed tabs.

Tunous:
I've changed the milestone selection dialog to consist of view pager that allows for selection of closed milestones.

I'm not 100% sure about the look of this. We could modify a bit the appearance of milestones to highlight which one is selected. And I'm not really sure if the "deselect" button is a good idea. Sadly I don't have a better one. Also, we might want to colorize the pager tabs.

What is missing is the functionality of old "Manage milestones" buttons. We could add it like in old dialog (perhaps stack buttons) but I'm open to other suggestions.

Preview:
milestone-dialog

Closes #570 

_From unrelated notes, I've added new wip label because I don't like adding WIP to titles._

### DIFF
--- a/app/src/main/java/com/gh4a/activities/IssueEditActivity.java
+++ b/app/src/main/java/com/gh4a/activities/IssueEditActivity.java
@@ -21,6 +21,7 @@ import android.content.Intent;
 import android.graphics.Typeface;
 import android.os.Bundle;
 import android.support.annotation.IdRes;
+import android.support.annotation.Nullable;
 import android.support.design.widget.AppBarLayout;
 import android.support.design.widget.CoordinatorLayout;
 import android.support.design.widget.FloatingActionButton;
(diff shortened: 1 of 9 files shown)


maniac103:
> I'm not 100% sure about the look of this. We could modify a bit the appearance of milestones to highlight which one is selected.

That would indeed make sense, I think.

> And I'm not really sure if the "deselect" button is a good idea. Sadly I don't have a better one.

What about placing a full-width 'Select none' or 'Any milestone' (wording improvements welcome) between the view pager and the button bar? That way we'd not mix selection and dialog dismissal in the button bar.

> Also, we might want to colorize the pager tabs.

Worth a try, but my gut feeling is that the current version will look better. In general I think having more than one large block of color (let alone different colors) looks 'too much'. It's basically the same thing as with the advanced editor.

> What is missing is the functionality of old "Manage milestones" buttons. We could add it like in old dialog (perhaps stack buttons) but I'm open to other suggestions.

What do you mean by that?

Tunous:
> > What is missing is the functionality of old "Manage milestones" buttons. We could add it like in old dialog (perhaps stack buttons) but I'm open to other suggestions.
>
> What do you mean by that?

Go to issue edit screen and click on the button to add milestone. You can see here that the dialog has additional button which is missing in new design.

maniac103:
> Go to issue edit screen and click on the button to add milestone. You can see here that the dialog has additional button which is missing in new design.

Ah, right. Should be easy to add though when moving the deselect button out.

Tunous:
Rebased onto latest master and finished missing changes. I'm still not convinced about the looks of this but at least now everything is functional.

PR #742, opened 2017-09-24
Improved filter visibility

Tunous:
Currently, when you enable any of the issue filters such as by labels or milestones there is no visual indicator that this filtering is in effect. I decided to fix this by adding subtitles to filter entries in right drawer and a "remove filter" button in the main toolbar. This solves few problems at once:

- Remove filter button gives a visual cue to the user that there is a filter in effect
- It also allows them to quickly return to default state
- Details in right drawer menu provide information about how exactly the filters are set up which makes it easier to see why some issues aren't appearing

Right now this is just a proof of concept implementation and it only includes milestones. I would like to hear what you think about this.

It relies on small hack: To allow menu items have subtitles I've overridden default `design_navigation_menu_item.xml` layout. An alternative solution would be to not use menu resources in the right drawer and implement custom RecyclerView.

Preview:
improved-filter


maniac103:
I think the small hack is fine and it's a very nice idea in general. Only thing that needs changes IMHO (except the missing implementation for other filters) is the icon: we use Google's filter icon (three horizontal lines of decreasing length) elsewhere, so the 'clear filter' icon should be consistent to that. Maybe we should use something in the style of the material version of [this icon](https://icons8.com/icon/19378/clear-filters) and use that icon (without cross, of course) to replace the existing filter icon?

smichel17:
I think I'd put the clear filter button to the right of the search button.

Tunous:
I used the icon from [here](https://materialdesignicons.com/icon/filter-remove). It also has variant without the cross. I don't know where to search for other icons that have both variants...

maniac103:
I'll draw new ones when I get some time. Shouldn't be too hard.

maniac103:
Should be fine now. @Tunous everything ok from your side?

Tunous:
Yeah, you've got it all. Merging.

PR #744, opened 2017-09-25
Improve pull requests status box

### LINKED ISSUE #738: Make CI status a link to build page on CI website
This is mildly related to #96 but should have a much smaller scope (so easier to implement) and be independent of the CI platform.

Currently OctoDroid shows the CI status on PRs, but it does not provide a way to actually open the CI website and check the build logs, etc.

It would be nice if there was a clickable link that would open the corresponding CI website in the user's browser at the build job's page, just like when you click here on the web UI:
image
Or here:
image

I don't know if the GitHub API exposes such links, but I guess it may.

Tunous:
# Improvements

- Colored status icon based on state (failing, pending, successful)
- Checks summary with counts and types
- Colored status icon for each check
- Collapsible status details, auto-expanded if there are pending or failing checks
- Clickable checks redirecting to browser (Closes #738)
- Blocked state support

# Notes

Right now the color for pending checks is same both in dark and light theme. It could be improved a bit in dark theme but I'm not sure how bright it should be. The current color is from GitHub website.

# Preview

Blocked with successful and failing checks:
blocked_checks_mixed

Blocked with successful checks:
blocked_checks_successful

Blocked without checks:
blocked_no_checks

Merge conflict with checks:
conflict_checks

Merge conflict without checks:
conflict_no_checks

Failing checks:
failing_checks

Pending checks:
pending_checks

Pending checks (dark theme):
pending_checks_dark

Pending and successful checks:
pending_checks_mixed

Successful checks:
successful_checks

Mergable without checks:
successful_no_checks


Tunous:
We could also swap the position of context and title for individual checks. On GitHub, the context is bigger so it would be nice to stay consistent with that.

app/src/main/java/com/gh4a/widget/CommitStatusBox.java
```diff
@@ -0,0 +1,211 @@
...
+        String pendingText = pendingCount > 0
+                ? getContext().getString(R.string.check_pending, pendingCount)
+                : "";
+        if ((failingCount > 0 || pendingCount > 0) && successCount > 0) {
+            pendingText += ", ";
+        }
+
+        String successfulText = successCount > 0
+                ? getContext().getString(R.string.check_successful, successCount)
+                : "";
```
maniac103: This may be overkill, but this seems like it's suited for using an ArrayList + TextUtils.join().

Tunous:
> This is why I went with using secondary color (replacing green + yellow) and primary color (replacing red) in the original design: the goal was to make the 'alarm' state stand out while still making it fit into the overall design.

I really disagree here. When I saw the primary color for first time I thought that it was a bug and not an intended look. It does bring user's attention but in a negative way. The read and green colors are better IMO. They also allow to very quickly distinguish the state of pull request. I'm just not happy with the color for pending state. This one probably can use secondary color...

maniac103:
My opinion isn't overly strong on this, so if yours is, just go ahead with this :-) But yeah, +1 for replacing the yellow by secondary color; there's no need for the pending status to stick out.

Tunous:
> We could also swap the position of context and title for individual checks. On GitHub, the context is bigger so it would be nice to stay consistent with that.

Thoughts on this?

Tunous:
I've discovered a new status and added support for it:

screenshot_20170927_131025
screenshot_1506510602

_Also removed colors for pending state_

Tunous:
I decided to do that context/description switch.

Another thing I noticed is that statuses are ordered differently on GitHub. Probably could change that here too. I'm just not sure what is the correct order (by status, successful last, with the same status ordered alphabetically based on context?)

maniac103:
> Another thing I noticed is that statuses are ordered differently on GitHub. Probably could change that here too. I'm just not sure what is the correct order (by status, successful last, with the same status ordered alphabetically based on context?)

Makes sense, that seems to match what GitHub is doing. Should probably do this in the loader and also remove the outdated statuses there.

PR #745, opened 2017-09-25
Display author association next to username

### LINKED ISSUE #690: Show user tags in PR/issue comments
Not sure if this data is available, but I think it'd be cool if we saw user tags in PR and issue comments.

By tags I'm referring to the text that appears at the top-right of comments in the web UI (only on the desktop UI, not mobile as far as I know) with `"owner"`, `"member"`, `"collaborator"`, `"contributor"`, etc.

Tunous:
author_association

Closes #690


Tunous:
> With the smaller font size and the font not being bold, one maybe even could get rid of the parentheses?

I also updated it to have secondary color for even less distraction but I prefer it with parentheses. That way it looks more like a tag instead of username suffix.

And extracted strings.

Tunous:
screenshot_1506431851

maniac103:
LGTM

PR #755, opened 2017-09-27
Change applicationId on debug builds

smichel17:
This allows installing debug and release builds side-by-side, so you can test WIP changes without uninstalling the current stable version.

When I've done this in the past, there's sometimes places where you'll get unintended dialogs prompting which app you want to complete the action in (for example, when both apps define the same intent filter). These can be fixed as we find them, and also they're acceptable since the people running debug versions know to expect that.

So far I have not done much testing, but I have started using this locally and it seems to work fine. The only duplicate intent filter I found so far is opening `gh4a://` links during github authorization. I don't have time to fix that right now, but I think this is ok to merge even so (or you can do it).

TODO:

- [x] Fix crash on opening any repo home page

### DIFF
--- a/app/src/main/java/com/gh4a/db/SuggestionsProvider.java
+++ b/app/src/main/java/com/gh4a/db/SuggestionsProvider.java
@@ -12,11 +12,16 @@ import android.provider.BaseColumns;
 import android.support.annotation.NonNull;
 import android.util.Log;
 
+import com.gh4a.BuildConfig;
+
+import java.util.Locale;
+
 public class SuggestionsProvider extends ContentProvider {
     private static final String TAG = "SuggestionsProvider";
 
     public interface Columns extends BaseColumns {
-        Uri CONTENT_URI = Uri.parse("content://com.gh4a.SuggestionsProvider/suggestions");
+        Uri CONTENT_URI = Uri.parse(String.format(Locale.US,
+                "content://%s.SuggestionsProvider/suggestions", BuildConfig.APPLICATION_ID));
 
         String TYPE = "type";
         String SUGGESTION = "suggestion";
@@ -33,7 +38,9 @@ public class SuggestionsProvider extends ContentProvider {
             sURIMatcher = new UriMatcher(UriMatcher.NO_MATCH);
 
     static {
-        sURIMatcher.addURI("com.gh4a.SuggestionsProvider", "suggestions", MATCH_ALL);
+        sURIMatcher.addURI(
+                String.format(Locale.US, "%s.SuggestionsProvider", BuildConfig.APPLICATION_ID),
+                "suggestions", MATCH_ALL);
     }
 
     private DbHelper mDbHelper;
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,6 +35,7 @@ android {
         targetSdkVersion 26
         versionCode 56
         versionName "4.2.2"
+        applicationId "com.gh4a"
 
         buildConfigField 'String', 'CLIENT_ID', ClientId
         buildConfigField 'String', 'CLIENT_SECRET', ClientSecret
@@ -42,6 +43,7 @@ android {
 
     buildTypes {
         debug {
+            applicationIdSuffix '.debug'
             zipAlignEnabled true
         }
         release {
--- a/app/src/debug/res/values/strings.xml
+++ b/app/src/debug/res/values/strings.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools"
+    tools:ignore="MissingTranslation">
+    <string name="app_name">OctoDroid Debug</string>
+</resources>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -157,12 +157,12 @@
 
         <provider
             android:name=".db.BookmarksProvider"
-            android:authorities="com.gh4a"
+            android:authorities="${applicationId}"
             android:exported="false" />
 
         <provider
             android:name=".db.SuggestionsProvider"
-            android:authorities="com.gh4a.SuggestionsProvider"
+            android:authorities="${applicationId}.SuggestionsProvider"
             android:exported="false" />
 
         <receiver
--- a/app/src/main/java/com/gh4a/db/BookmarksProvider.java
+++ b/app/src/main/java/com/gh4a/db/BookmarksProvider.java
@@ -16,12 +16,16 @@ import android.util.Log;
 import android.widget.Toast;
 
 import com.gh4a.R;
+import com.gh4a.BuildConfig;
+
+import java.util.Locale;
 
 public class BookmarksProvider extends ContentProvider {
     private static final String TAG = "BookmarksProvider";
 
     public interface Columns extends BaseColumns {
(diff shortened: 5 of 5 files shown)


maniac103:
Fixing that would require putting the URL scheme into a gradle variable and using that in the manifest and code, I guess?

smichel17:
It might be do-able with a separate manifest file instead (like with disabling analytics in the foss variant), I can check later tonight

WanderMax:
crash when tap on comments

smichel17:
Started looking into the crash. It's is a permission issue related to the bookmarks provider.

@maniac103 How do you want to handle gh4a:// links? Something different for the debug variant or just leave them both to registered?

@WanderMax If it happens in the original debug build then it's not related to this PR and I'm not going to try to fix it here; I'd report it somewhere else so it doesn't get lost.

maniac103:
@smichel17 Depends on how complicated the fix is. If it's complicated or ugly to fix, I'd just leave it as is; if the fix is easy I'd fix it.
(BTW, it isn't manifest alone, the activity IIRC also checks the scheme)

WanderMax:
`Fix crash on opening any repo home page ` what's the home page?

smichel17:
https://github.com/slapperwan/gh4a

Actually it is any page that interacts with bookmarks.

WanderMax:
I open the https://github.com/slapperwan/gh4a page via OctoDroid from mail.
It seems OK, nothing crashes.

smichel17:
Are you using a debug build off of the branch of this pull request?

WanderMax:
I make little modification based on your commit, it's strange that I have no crash when open GitHub link via my build.

smichel17:
@WanderMax What app name did you use?

smichel17:
@maniac103 The crash was due to the scheme being defined statically in the providers, too; thanks for the tip. Fixed the content providers and rebased on the latest master. 

As for the oauth scheme: on the code side, it is not difficult. However, on github's side, there's only one authorization callback uri per oath app, so you'd have to create a separate one for the debug version. So, your call.

WanderMax:
@smichel17  Here is my build.gradle as below:
`def taskRequests = getGradle().getStartParameter().getTaskRequests().toString()
def isPlaystore = taskRequests.contains("Playstore") || taskRequests.contains("playstore")

buildscript {
    repositories {
        maven { url 'https://maven.fabric.io/public' }
    }

    dependencies {
        classpath 'io.fabric.tools:gradle:1.+'
    }
}

repositories {
    mavenCentral()
    maven { url 'https://maven.fabric.io/public' }
    maven { url "https://jitpack.io" }
    maven { url "https://maven.google.com" }
}

apply plugin: 'com.android.application'

def releaseTime() {
    return new Date().format("yyyyMMdd_HHmm", TimeZone.getTimeZone("GMT+08"))
}

if (isPlaystore) {
    apply plugin: 'io.fabric'
}

android {
    signingConfigs {
        playStore
    }
    compileSdkVersion 26
    buildToolsVersion '26.0.1'
    defaultConfig {
        minSdkVersion 21
        targetSdkVersion 26
        applicationId "com.gh4a"
        versionCode 56
        versionName "4.2.2"
        ndk {
            abiFilters "armeabi", "armeabi-v7a", "arm64-v8a"
        }

        buildConfigField 'String', 'CLIENT_ID', ClientId
        buildConfigField 'String', 'CLIENT_SECRET', ClientSecret
    }

    buildTypes {
        debug {
            zipAlignEnabled true
            applicationIdSuffix '.pro'
            applicationVariants.all { variant ->
                variant.outputs.each { output ->
                    def outputFile = output.outputFile
                    if (outputFile != null && outputFile.name.endsWith('.apk')) {
                        // Set apk name to OctoHub Pro_v0.1(233)_20170226_1633.apk
                        def fileName = "OctoHub_Pro_v${defaultConfig.versionName}(${defaultConfig.versionCode})_${releaseTime()}.apk"
                        output.outputFile = new File(outputFile.parent, fileName)
                    }
                }
            }
        }
        release {
            zipAlignEnabled true
            signingConfig signingConfigs.playStore
        }
    }

    productFlavors {
        // includes proprietary libs
        playstore {

        }

        // only foss
        foss {

        }
    }
}

dependencies {
    compile project(':github-api')
    compile 'com.android.support:support-v4:26.0.1'
    compile 'com.android.support:appcompat-v7:26.0.1'
    compile 'com.android.support:cardview-v7:26.0.1'
    compile 'com.android.support:preference-v7:26.0.1'
    compile 'com.android.support:preference-v14:26.0.1'
    compile 'com.android.support:customtabs:26.0.1'
    compile 'com.android.support:design:26.0.1'
    compile 'com.larswerkman:HoloColorPicker:1.5@aar'
    compile 'com.caverock:androidsvg:1.2.2-beta-1@aar'
    compile 'pl.droidsonroids.gif:android-gif-drawable:1.1.+'
    compile 'org.ocpsoft.prettytime:prettytime:4.0.1.Final'
    compile 'com.github.castorflex.smoothprogressbar:library:1.1.0'
    compile 'org.ccil.cowan.tagsoup:tagsoup:1.2.1'
    compile 'com.github.pluscubed:recycler-fast-scroll:0.3.2@aar'
    compile('com.vdurmont:emoji-java:3.2.0') {
        exclude group: 'org.json', module : 'json'
    }
    compile 'com.github.Tunous:MarkdownEdit:0.1.1'
    playstoreCompile('com.crashlytics.sdk.android:crashlytics:2.6.5@aar') {
        transitive = true;
    }
    testCompile 'junit:junit:4.12'
    testCompile 'org.robolectric:robolectric:3.4.2'
}

def Properties props = new Properties()
def propFile = new File('signing.properties')
if (propFile.canRead()) {
    props.load(new FileInputStream(propFile))
}

if (props.containsKey('STORE_FILE')
        && props.containsKey('STORE_PASSWORD')
        && props.containsKey('KEY_ALIAS')
        && props.containsKey('KEY_PASSWORD')) {
    android.signingConfigs.playStore.storeFile = file(props['STORE_FILE'])
    android.signingConfigs.playStore.storePassword = props['STORE_PASSWORD']
    android.signingConfigs.playStore.keyAlias = props['KEY_ALIAS']
    android.signingConfigs.playStore.keyPassword = props['KEY_PASSWORD']
} else {
    println 'signing.properties not found or incomplete'
    android.buildTypes.release.signingConfig = null
}
`

smichel17:
@maniac103 @Tunous What's needed to move forward with this?

app/src/main/java/com/gh4a/db/SuggestionsProvider.java
```diff
@@ -12,11 +12,14 @@
...
 
+import com.gh4a.BuildConfig;
+
 public class SuggestionsProvider extends ContentProvider {
     private static final String TAG = "SuggestionsProvider";
 
     public interface Columns extends BaseColumns {
-        Uri CONTENT_URI = Uri.parse("content://com.gh4a.SuggestionsProvider/suggestions");
+        Uri CONTENT_URI = Uri.parse(String.format("content://%s.SuggestionsProvider/suggestions",
+                                                  BuildConfig.APPLICATION_ID));
```
Tunous: Code style: As far as I remember we use 8 spaces as continuation indent instead of aligning with first parameter.

For me personally code aligned like you've done it here is harder to read because of less available space.  

This applies to other places too.

PR #756, opened 2017-09-27
remove issue/pr reaction menu item

smichel17:
You can already react to an issue/pr by reacting to the top (creation) comment. That makes the `Add reaction` menu item a duplicate function that's just taking up space. This removes it.

I'm not 100% sure that I've nixed all the associated code, but I was cautious of breaking reactions elsewhere in the page.


maniac103:
The top (creation) comment doesn't have an overflow menu though. Instead, it has that functionality added to the toolbar menu ;-)

smichel17:
Ah, you are right. And I am right. If someone has reacted to the top comment, then there is a menu down at the bottom to add additional reactions

smichel17:
@maniac103 I think the solution is to remove the menu item iff there are already reactions on the post. That's slightly more work, but I'll look into it at some point if you'd accept a PR for it.

Either way, it won't happen right away and this PR should not be merged as-is, so closing for now and I'll re-open or start a new PR later.

maniac103:
I'm not entirely sure whether that's a good idea. On one hand it indeed avoids redundancy in the menus, on the other hand it might confuse users who already have found the menu. In general I tend to not mess with hiding/removing menu items if not absolutely necessary.
@Tunous Do you have an opinion on this?

Tunous:
Yes, I think that it's a bad idea too.

smichel17:
Okay, easier option for me :)

smichel17:
Example image

I just noticed the gitlab web interface does this in a way that would avoid duplication and an annoying sub-sub-menu just to add a reaction: always show a subset of reactions for the top comment, even if no-one has reacted yet.

@Tunous @maniac103 Would you be open to doing it this way?

smichel17:
I think I did not explain well.

screenshot-github

Github web has a react icon in line with the username, for both issues/prs and comments. When an issue/pr or comment has a reaction, it shows a reaction bar at the bottom, too. Its reaction bar does not have a menu to add new reactions.

screenshot-fasthub

Fasthub does the same.

screenshot-octodroid

Octodroid always hides reactions in sub-menus. For issues/prs, reactions are particularly far away from the content, which doesn't make sense to me.



When an issue/pr or comment has a reaction, OctoDroid shows a reaction bar below the issue, which contains an additional reaction menu.

screenshot-gitlab

For comments, Gitlab web has the same initial reaction menu as github and fasthub, and when there are reactions, adds a reaction bar w/menu, like OctoDroid.

For issues/prs, it does not have the react menu at the top, but rather always shows the reaction bar w/menu at the bottom.

---

I think that for comments, OctoDroid should do it the same way as everyone else: a reaction menu that sits next to the overflow menu. 

For issues/prs, OctoDroid cannot do it the same way, because the reaction menu icon would be too close to the edit FAB. So, I suggest we do it like Gitlab web does: always show the reaction bar.

I agree that always showing reactions for comments would be a ridiculous waste of space. I don't think it is a big waste for issues/prs (particularly since issues/prs often have reactions anyway, so the reaction bar is often shown, anyway).

---

In terms of implementing, that leaves:

- [x] Remove the current reaction menu for issues/prs
- [ ] Always show the reaction bar for issues/prs
- [ ] Add an icon for the reaction menu in comments
- [ ] Remove the reaction menu from the overflow menu
- [ ] Add a reaction menu for comments that sits next to the overflow menu

maniac103:
I understood you well, I just don't agree with showing the reaction bar permanently - handling issue and comment differently feels inconsistently to me. That's why I suggested to
- show the reaction button (for both comments and issues) like FastHub does (but smaller, like our overflow button) iff no reactions exist [button would open the popup menu that's currently accessed as submenu]
- do *not* show that button and instead show reaction bar iff reactions exist

As mentioned, I realize that there's a potential conflict of touch targets, but I think it can be solved with adding a little top margin to the reaction button, given that user avatar + name + timestamp will be larger than that button.

smichel17:
> show the reaction button (for both comments and issues) like FastHub does (but smaller, like our overflow button)

Okay. And of course we can always reevaluate if if it turns out that the touch targets do conflict. 

> do not show that button and instead show reaction bar iff reactions exist

I don't think this part is necessary. Unless you also want to remove the reaction submenu on comments when the reaction bar is shown?

maniac103:
> Okay. And of course we can always reevaluate if if it turns out that the touch targets do conflict.

Right. Best idea, I think, would be doing a mockup (only add a nonfunctional button) first to see whether a meaningful placement can be found, then make it functional.

> I don't think this part is necessary. Unless you also want to remove the reaction submenu on comments when the reaction bar is shown?

Actually I want to completely remove the submenu in favor of reaction button and reaction bar, respectively.

Issue #323, opened 2016-05-21
Add commit signature indicator


maniac103:
Github has started verifying commit PGP signatures. We should add an indicator for signed commits.

https://developer.github.com/changes/2016-04-04-git-signing-api-preview/

Tunous:
Out of preview: https://developer.github.com/changes/2017-10-12-git-signing/

PR #778, opened 2017-10-16
Rx refactor v2

Tunous:


app/src/main/java/com/gh4a/utils/RxUtils.java
```diff
@@ -0,0 +1,185 @@
...
+import io.reactivex.SingleTransformer;
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.functions.Function;
+import io.reactivex.functions.Predicate;
+import io.reactivex.processors.PublishProcessor;
+import io.reactivex.schedulers.Schedulers;
+import retrofit2.Response;
+
+public class RxUtils {
+    public static <T> Single<T> throwOnFailure(Single<Response<T>> upstream) {
```
Tunous: I think that this method is useless. Can't we simply call `.map(ApiHelpers::throwOnFailure) `instead of `.compose(RxUtils::throwOnFailure)`?

app/src/main/java/com/gh4a/utils/RxUtils.java
```diff
@@ -0,0 +1,185 @@
...
+                if (((ApiRequestException) error).getStatus() == code) {
+                    return Single.just(value);
+                }
+            }
+            return Single.error(error);
+        });
+    }
+
+    public static <T> Single<T> doInBackground(Single<T> upstream) {
+        return upstream.subscribeOn(Schedulers.newThread())
```
Tunous: Might be better to use `Schedulers.io()` or `Schedulers.computation()` to not create new thread every time.

app/src/main/java/com/gh4a/BaseActivity.java
```diff
@@ -112,6 +112,8 @@
     private Animator mHeaderTransition;
     private final Handler mHandler = new Handler();
 
+    private final List<Disposable> mDisposeOnStop = new ArrayList<>();
```
Tunous: Use `CompositeDisposable` and its `dispose` method.

Tunous:
And I've found [this library](https://github.com/T-Spoon/Traceur) which might solve the problem with stack traces you've mentioned @maniac103

app/src/main/java/com/gh4a/utils/RxUtils.java
maniac103: That one at least partly was intentional: we launched the AsyncTasks on the parallel executor before. Not sure what exactly AsyncTaskLoader does.

app/src/main/java/com/gh4a/BaseActivity.java
maniac103: Is that one clearable, IOW can one reuse it after calling dispose?

app/src/main/java/com/gh4a/utils/RxUtils.java
maniac103: Will check this.

app/src/main/java/com/gh4a/BaseActivity.java
Tunous: Apparently you can no longer use it after calling `dispose`. But it might be possible to use clear `instead` which based on [documentation](http://reactivex.io/RxJava/javadoc/io/reactivex/disposables/CompositeDisposable.html#clear--) should dispose items.

app/src/main/java/com/gh4a/activities/home/HomeActivity.java
```diff
@@ -524,6 +505,32 @@ private void switchTo(int itemId, FragmentFactory factory) {
...
+        UserService service = Gh4Application.get().getGitHubService(UserService.class);
+        service.getUser(mUserLogin)
+                .map(ApiHelpers::throwOnFailure)
+                .compose(makeLoaderSingle(ID_LOADER_USER, force))
+                .subscribe(result -> {
+                    Gh4Application.get().setCurrentAccountInfo(result);
+                    mUserInfo = result;
+                    updateUserInfo();
+                }, error -> {});
+    };
```
Tunous: Trailing semicolon

app/src/main/java/com/gh4a/activities/EditCommitCommentActivity.java
```diff
@@ -4,35 +4,42 @@
...
+    protected Single<GitHubCommentBase> createComment(String repoOwner, String repoName,
+            String body, long replyToCommentId) {
+        RepositoryCommentService service =
+                Gh4Application.get().getGitHubService(RepositoryCommentService.class);
+        CreateCommitComment request = CreateCommitComment.builder().body(body).build();
         String sha = getIntent().getStringExtra("commit");
-        commitService.addComment(repoId, sha, comment);
+        return service.createCommitComment(repoOwner, repoName, sha, request)
+                .map(ApiHelpers::throwOnFailure)
+                .map(response -> response);
```
Tunous: What's the point of this map?

app/src/main/java/com/gh4a/activities/FileViewerActivity.java
```diff
@@ -343,4 +306,38 @@ private static String highlightImage(String imageUrl, String cssTheme, String ti
...
+                .onErrorResumeNext(error -> {
+                    if (error instanceof ApiRequestException) {
+                        ClientErrorResponse response = ((ApiRequestException) error).getResponse();
+                        List<ClientErrorResponse.FieldError> errors =
+                                response != null ? response.errors() : null;
+                        if (errors != null) {
+                            for (ClientErrorResponse.FieldError fe : errors) {
+                                if (fe.reason() == ClientErrorResponse.FieldError.Reason.TooLarge) {
+                                    openUnsuitableFileAndFinish();
+                                    return Single.just(null);
```
Tunous: `Single.just` doesn't accept null as value. We might want to use `Maybe.empty()` instead, though that will require some more changes to work with `Maybe`.

app/src/main/java/com/gh4a/utils/RxUtils.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    public static <T> Single<Response<Page<T>>> searchPageAdapter(Single<Response<SearchPage<T>>> upstream) {
+        return upstream.map(response -> {
+            if (response.isSuccessful()) {
+                return Response.success(new ApiHelpers.SearchPageAdapter<>(response.body()));
+            }
+            return Response.error(response.errorBody(), response.raw());
+        });
+    };
```
Tunous: Trailing semicolon

app/src/main/java/com/gh4a/utils/RxUtils.java
```diff
@@ -0,0 +1,194 @@
...
+            private void hideDialog() {
+                if (mFragment.getActivity() != null) {
+                    mFragment.dismissAllowingStateLoss();
+                }
+                mFragment = null;
+            }
+        };
+    }
+
+    public static <T> SingleTransformer<T, T> wrapWithRetrySnackbar(final FragmentActivity activity,
```
Tunous: Unused parameter: 'activity'

app/src/main/java/com/gh4a/activities/EditCommitCommentActivity.java
maniac103: Making the type system happy ;-)
Here: convert from GitComment to GitHubCommentBase.

Tunous: And what makes it unhappy? If I remove that line I don't see any warnings and the app still compiles.

maniac103: It didn't compile when I introduced this. It said that a Single<GitComment> isn't compatible to the supposed return type Single<GitHubCommentBase>.

I'll check this again.

Tunous:
Starring and watching is incorrect. When I press on watch button it toggles starred status. Pressing on star does nothing (or marks repo as watched?)

maniac103:
Indeed, starring and watching is mixed up: https://github.com/slapperwan/gh4a/blob/rx/app/src/main/java/com/gh4a/activities/RepositoryActivity.java#L306

Tunous:
- [ ] Crash when creating new issue
- [ ] Refreshing in repository screen locks About tab in loading state

maniac103:
> Crash when creating new issue

Damn, I have a feeling we'll hit more of those subtle differences in the bindings. The old bindings initialized missing ints (by default with 0), with the new bindings they're null.

maniac103:
> After adding new comment or editing existing loading bar doesn't appear. Instead for a short while all comments are replaced with blank space

@Tunous Which kind of comment do you mean here?

Tunous:
Issue/pr comments

Tunous:
Another issue:

- [ ] Some gists don't load

maniac103:
Adding issue comments seems to work for me.
Please define 'some' in 'some gists' ;-)

Tunous:
> Adding issue comments seems to work for me.

Yeah looks like it's fixed now.

- [ ] Code search doesn't show matched code blocks
- [ ] Crashed after deleting a comment. Might be related to next problem:
- [ ] Tunous/TestingRepo#43 crashes due to `comment_deleted` event. We should add special handling for this type of errors to avoid this issue in future. It should log the error (preferably to crashlytics) but not crash the app.

I found a list of events in [GraphQL API Explorer](https://developer.github.com/v4/explorer/). Search for "event" in the documentation sidebar.

Tunous:
Not sure what happened but when I opened this pull request the activity crashed. It worked the second time though...

maniac103:
> Not sure what happened but when I opened this pull request the activity crashed. It worked the second time though...

99.9% likelihood of this being due to the 'mergeable' field missing in the PR response (or it being lost in the JSON-to-POJO conversion). I've seen this sporadically on different PRs, but am completely stumped as to why it happens. I guess I'll have to add log output for this and/or do a test like querying for a given PR in a loop in order to see whether it's server side or in the conversion logic. It could be the former - with the old bindings, such a problem would simply be left unnoticed as the value would be initialized (to false).

> Crash when clicking on reviewed file of a review (above "Show details" button)

```
10-30 20:25:39.478 E/AndroidRuntime(13903): java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
10-30 20:25:39.478 E/AndroidRuntime(13903): 	at com.gh4a.adapter.timeline.ReviewViewHolder.onClick(ReviewViewHolder.java:270)
```

Looks like a trivial fix :-)

maniac103:
- [ ] Clicking on a notification doesn't (visually) mark it as read

maniac103:
I'm nearing completion here; I think it's in a state where broader testing would be useful. As the rebasing is pretty painful, I'd like to merge to master soon.

@Tunous What are your thoughts about merging back, and what PRs should we merge before doing so?

Tunous:
> What are your thoughts about merging back

Should be good. Just give it a day or two so I can do one final test run on my main device.

> what PRs should we merge before doing so?

Mostly we have PRs for new features so they don't need to be merged before this. Perhaps only #753 would be nice to have in next update as it fixes an issue with comments not refreshing. But need to finish it first though.

___

One more issue I found:

- List of issues doesn't load more pages (e.g. for this repo)

maniac103:
> List of issues doesn't load more pages (e.g. for this repo)

This was an SDK issue: https://github.com/maniac103/GitHubSdk/commit/5db6638b9b2d407e5982c1a4882266f48a84115c

maniac103:
- [ ] Switching context in news feed only works once

Tunous:
IDK if related but [this pr](https://github.com/unhappychoice/Norimaki/pull/26) doesn't load.

maniac103:
> IDK if related but this pr doesn't load.

Very much related, a user type is missing in the enum:
```
11-05 18:16:03.494 D/Gh4a    (25346): com.squareup.moshi.JsonDataException: Expected one of [User, Organization] but was Bot at path $.items[0].user.type
```

Tunous:
BTW we could display messages like this in error view. At least in debug builds.

Tunous:
It looks like label removed events do not appear. See #323

maniac103:
I don't understand - how are label events related to commit signature support?

Tunous:
I mean, in this issue I don't see the label removed event. View it in browser - there that event appears.

maniac103:
Fixed mentioned issues and rebased onto master; cleaned up history a bit during the rebase (e.g. made every commit compilable against the latest SDK bindings).

maniac103:
- [ ] Deleting PR branches doesn't work (likely an SDK issue)

Tunous:
- [ ] Statuses have incorrect type for `id` should be `Long` instead of `Integer`. [Happens here](https://github.com/square/retrofit/pull/2550)

```
11-08 14:13:50.912 com.gh4a D/OkHttp: --> GET https://api.github.com/repos/square/retrofit/commits/ffe64f6888873ce438798c3db11f304df45f23ca/statuses?page=1 http/1.1
11-08 14:13:51.238 com.gh4a D/OkHttp: <-- 200 OK https://api.github.com/repos/square/retrofit/commits/ffe64f6888873ce438798c3db11f304df45f23ca/statuses?page=1 (325ms, 8071-byte body)
11-08 14:13:51.245 com.gh4a D/RxAndroidLoader: delivering result
11-08 14:13:51.253 com.gh4a D/Gh4a: Loading data failed
                                    com.squareup.moshi.JsonDataException: Expected an int but was 4303719446 at path $.items[0].id
```

Also, there appear to be multiple requests for the same thing when viewing pull requests. For example, I can see the same request for reactions, files, comments, etc twice. Idk if it's an issue with logging or some other error but we should look at this. I also checked it when going to individual issues, list of them, stargazers, forks and the duplication didn't happen here, just happens to individual pull requests.

maniac103:
> Statuses have incorrect type for id should be Long instead of Integer.

Hmm, I wonder whether that's the case for statuses only or for *all* ID fields...

Tunous:
True, I'd bet that it's indeed all (or almost all) of them.

maniac103:
> Also, there appear to be multiple requests for the same thing when viewing pull requests. For example, I can see the same request for reactions, files, comments, etc twice. Idk if it's an issue with logging or some other error but we should look at this. I also checked it when going to individual issues, list of them, stargazers, forks and the duplication didn't happen here, just happens to individual pull requests.

It's a matter of how the timeline is constructed in PullRequestFragment::onCreateDataSingle:
- comments are fetched there and in PullRequestFilesFragment
- files are fetched there and in PullRequestFilesFragment
- reviews are fetched there and in PullRequestActivity (for determining pending review)

Only interesting thing is the issue reactions; that fetch should happen only once.

maniac103:
- [ ] We should cache responses (fortunately extremely simply with OkHttp, and doing so will automatically enable ETag handling)

Tunous:
- [ ] Crashes when searching for repositories

maniac103:
> Crashes when searching for repositories

Not for me seemingly? What does the stacktrace look like?

maniac103:
- [ ] NPE in PullRequestActivity.java:235 after submitting a review (rebase remainder)

Tunous:
> Not for me seemingly? What does the stacktrace look like?

Can check that tomorrow when I'll be at my PC. But now I see that it doesn't happen for every search terms. I tested it with "develop.github.com"

maniac103:
```
11-08 21:27:00.685 E/AndroidRuntime(3498): io.reactivex.exceptions.OnErrorNotImplementedException: API rate limit exceeded for maniac103. (403)
```

maniac103:
That caching layer alone is worth the conversion to the new bindings - it's incredible how fast everything gets after the initial fetch.

maniac103:
I asked support about it.

Tunous:
- [ ] Cache works incorrectly for notifications. If you mark the notification as read, close the app and return to notifications screen it'll display the cached data with that notification still marked as unread. Only doing pull to refresh action loads new data.

maniac103:
> Cache works incorrectly for notifications. If you mark the notification as read, close the app and return to notifications screen it'll display the cached data with that notification still marked as unread. Only doing pull to refresh action loads new data.

That's intended by Github though. They send a `max-age=60` Cache-Control header, which allows caching the response for 60 seconds, which is exactly what OkHttp is doing - after those 60 seconds, it falls back to using ETags.
Evicting single URLs from the OkHttp cache is not possible, and clearing the whole cache (e.g. on application object destroy) seems a little heavy handed. I wonder whether we could hide the max-age header in the response from the caching layer by using an interceptor...

Tunous:
> but avoiding the network latency is a good thing when quickly going back-and-forth, isn't it?

Yeah, it's good but in this case, it looks more like a bug. At first, I thought that marking as read was broken but it turned out that the app was simply loading the same notification from the cache (and there was a more recent unread notification that didn't get loaded!). It's definitely not something that we want users to experience.

maniac103:
I'll play with the interceptor; it should be totally possible to nuke the header (and thus use ETags only).

maniac103:
I added the interceptor, but notifications seemingly still are off. I'll check this.

maniac103:
- [ ] Crash when trying to open account chooser

Tunous:
> it looks like they don't assign a new ETag when marking notifications as read?

Maybe the call needs `since` parameter? Otherwise it would always default to current time which is different every time. Though I'm not sure how that would work, just a random thought.

maniac103:
> Maybe the call needs since parameter? Otherwise it would always default to current time which is different every time

I believe their docs are wrong there, as 'give me all notifications updated after <now>' doesn't make sense and would always return an empty list. IMHO `since` rather defaults to 0 ('beginning of time'). I'll repro and ask them.

maniac103:
This definitely looks like a server side bug. They send `ETag: ""` on every response. IMHO, they should either send unique ETags or none at all. I'll ask them.

Tunous:
- [ ] Crash when trying to open issue creation screen

Tunous:
- [ ] Repository search for "kotlin" crashes app. (Not on PC so I don't have stacktrace)

maniac103:
> Repository search for "kotlin" crashes app.

Seemingly not for my. I only get an 'API rate limit exceeded' error

(We should probably make SearchFragment use paging)

PR #784, opened 2017-10-24
Detach feature

### LINKED ISSUE #710: Issue detail screen: enhancement

Tunous:
This is another approach to a feature implemented in #715. This implementation makes it optionally available through "Detach" button in the toolbar menu instead of being launched automatically. That way we get all the benefits of this feature without confusing users with changed launching behavior.

# Behavior

detach-task

- When the app takes the full screen the new task also takes full screen and is brought to the front.
- When the app is in split-screen then the detached task will appear in the second split.

# Further improvements

- [ ] Decide which screens should include this detach button
  - [x] Issue
  - [x] Pull request
  - [x] File view
  - [x] Diff view
  - [ ] Repository
  - [x] User
  - [x] Commit
  - [ ] More?

- [ ] Possibly remove the back button from the toolbar for the detached task?
I'm not sure about this as it removes the possibility to navigate up through the hierarchy in the new task. But it would make it clear that this specific task is supposed to represent a single issue. With it, we could also update the title for the entry in the recent tasks list to mention number/name of the issue instead of generic "OctoDroid".

Closes #710 and #715.


smichel17:
This implementation looks super nice. 

> Possibly remove the back button from the toolbar for the detached task?
I'm not sure about this as it removes the possibility to navigate up through the hierarchy in the new task. But it would make it clear that this specific task is supposed to represent a single issue. With it, we could also update the title for the entry in the recent tasks list to mention number/name of the issue instead of generic "OctoDroid".

I see two ways to go with this:

1. **Pop out this page only.** In this case, we would remove the back button (actually, replace it with an X to close the task) and include the number/name in the title of the task. 
2. **Pop out the entire repo.** In this case, we would include the back button, up until the level of the repo, which would not have a back button. The task title would be the name of the repo.

app/src/main/java/com/gh4a/utils/TaskUtils.java
```diff
@@ -0,0 +1,23 @@
...
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP)
+    public static void startNewTask(Context context, Intent intent) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_DOCUMENT);
+            intent.addFlags(Intent.FLAG_ACTIVITY_MULTIPLE_TASK);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                intent.addFlags(Intent.FLAG_ACTIVITY_LAUNCH_ADJACENT);
+            }
+        }
```
maniac103: Did you try what happens on API < 20 here? Also, no FLAG_ACTIVITY_NEW_TASK?

app/src/main/res/values/strings.xml
```diff
@@ -547,4 +547,5 @@
     <string name="notification_subject_pr">PR</string>
     <string name="notification_subject_release">Release</string>
     <string name="notification_interval">Sync interval</string>
+    <string name="detach">Detach</string>
```
maniac103: Maybe 'Open in new window' or similar? 'Detach' may be a bit too technical.

app/src/main/java/com/gh4a/activities/IssueActivity.java
```diff
@@ -73,6 +74,12 @@ public static Intent makeIntent(Context context, String login, String repoName,
                 .putExtra("initial_comment", initialComment);
     }
 
+    public static void startTask(Context context, String login, String repoName, int number,
+            IntentUtils.InitialCommentMarker initialComment) {
+        Intent intent = makeIntent(context, login, repoName, number, initialComment);
+        TaskUtils.startNewTask(context, intent);
+    }
```
maniac103: Can't you inline this below?

maniac103:
> I see two ways to go with this:
> 
> 1. Pop out this page only. In this case, we would remove the back button (actually, replace it with an X to close the task) and include the number/name in the title of the task.
> 2. Pop out the entire repo. In this case, we would include the back button, up until the level of the repo, which would not have a back button. The task title would be the name of the repo.

Option 1 would make more sense to me, given the menu option is in the overflow menu of a specific page.

app/src/main/java/com/gh4a/utils/TaskUtils.java
Tunous: Yeah I've tested it previously. As far as I remember older versions didn't have tasks feature so it always opened in the same task. For them we can remove that action from menu. I'm not sure about KitKat though. Will check again later.

Tunous:
Disabled this feature for versions earlier than L because it doesn't work here.

Also removed the up button for this new task but it's a bit problematic. An example situation.

1. Enter issue from default task
2. Start a new task 
It won't have up button
3. Go to issue edit screen
4. Press on the up arrow
We return to the previous screen but now up button appears

This should be fixed somehow so the up button doesn't reappear but I have no ideas...

smichel17:
For option 1, I think I'd start with issue, PR, diff, and file view (given, the last one is awkward since it's normally a tab; maybe start with just issue and PR and see whether we miss not being able to navigate or pop out other parts

Tunous:
I've refactored initialization a bit so now it's possible to add this detach feature by overriding single method. And added it to more screens.

app/src/main/java/com/gh4a/utils/TaskUtils.java
```diff
@@ -0,0 +1,31 @@
...
+    public static void startNewTask(@NonNull Context context, @NonNull Intent intent) {
+        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        intent.addFlags(Intent.FLAG_ACTIVITY_NEW_DOCUMENT);
+        intent.addFlags(Intent.FLAG_ACTIVITY_MULTIPLE_TASK);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            intent.addFlags(Intent.FLAG_ACTIVITY_LAUNCH_ADJACENT);
+        }
+        intent.putExtra(EXTRA_NEW_TASK, true);
+        context.startActivity(intent);
+    }
```
maniac103: Can't we put those two methods into IntentUtils, given they deal with intents anyway?

Tunous: Done.

maniac103:
Should probably hide the 'open in new window' menu item if the activity has the 'new task' extra.

PR #790, opened 2017-10-29
Adding support for PullRequest Reviews

edyesed:
My org uses Reviews for PullRequests on the regular. 

This client is great, and it didn't have support for Reviews. 

Now it does. 

I'm not an Android nor Java developer ( as you are likely to realize as you review this PR ).

Anyway, I hope you'll accept this PR. When you go to any PR you'll notice a "Review Changes" button at the top right of the window. That opens an editor popup where you can Approve, Request Changes, or Comment as a Review on a PR. 

I wish I knew enough java/android to add to the test suite, but I do not.


app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -246,6 +289,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
             case R.id.pull_merge:
                 showMergeDialog();
                 break;
+            case R.id.pull_review:
+                /* come back and do something on approve */
```
Tunous: What does this comment mean?

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -90,6 +94,7 @@ public static Intent makeIntent(Context context, String repoOwner, String repoNa
     public static final int PAGE_CONVERSATION = 0;
     public static final int PAGE_COMMITS = 1;
     public static final int PAGE_FILES = 2;
+    public static final int PAGE_REVIEWS = 3;
```
Tunous: Unused?

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -143,6 +167,24 @@ protected void onResultReady(PullRequest result) {
         }
     };
 
+
+    private final LoaderCallbacks<Review> mReviewCallback = new LoaderCallbacks<Review>(this) {
```
Tunous: Unused?

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -614,4 +711,43 @@ protected String getErrorMessage() {
...
+        protected ProgressDialogTask<Review> clone() {
+            return new PullRequestReviewTask(mReviewMessage, mReviewEvent);
+        }
+
+        @Override
+        protected Review run() throws Exception {
+            PullRequestService pullService = (PullRequestService)
+                    Gh4Application.get().getService(Gh4Application.PULL_SERVICE);
+            RepositoryId repoId = new RepositoryId(mRepoOwner, mRepoName);
+            mReview = pullService.createPullRequestReview(repoId, mPullRequest.getNumber(), mReviewEvent, mReviewMessage);
```
Tunous: Should simply return the result of this method call instead of assigning it to a field.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -614,4 +711,43 @@ protected String getErrorMessage() {
...
+            return mReview;
+        }
+
+        /* There's nothing to be done on success */
+        @Override
+        protected void onSuccess(Review result) {
+            mReview = result;
```
Tunous: We should reload all comments so the newly created review will be added to the screen. Also the `mReview` field appears to be unused so you can remove it.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -614,4 +711,43 @@ protected String getErrorMessage() {
...
+        @Override
+        protected Review run() throws Exception {
+            PullRequestService pullService = (PullRequestService)
+                    Gh4Application.get().getService(Gh4Application.PULL_SERVICE);
+            RepositoryId repoId = new RepositoryId(mRepoOwner, mRepoName);
+            mReview = pullService.createPullRequestReview(repoId, mPullRequest.getNumber(), mReviewEvent, mReviewMessage);
+            return mReview;
+        }
+
+        /* There's nothing to be done on success */
```
Tunous: This comment won't be needed.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -614,4 +711,43 @@ protected String getErrorMessage() {
...
+        @Override
+        protected void onSuccess(Review result) {
+            mReview = result;
+        }
+
+        @Override
+        protected String getErrorMessage() {
+            /* How do you get at the response/result of pullService.createPullRequestReview ?? */
```
Tunous: The result is passed to `onSuccess` method above.

app/src/main/res/layout/pull_review_message_dialog.xml
```diff
@@ -0,0 +1,36 @@
...
+    <com.gh4a.widget.StyleableTextView
+        android:id="@+id/notice"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
+        android:text="@string/pull_request_review_description"
+        android:textAppearance="?android:attr/textAppearanceSmall" />
+
+    <EditText
+        android:id="@+id/et_commit_message"
```
Tunous: Needs some text options like auto-capitalized sentences, autocorrection, etc. Look at how it's done in the input field for comments.

app/src/main/res/layout/pull_review_message_dialog.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:cursorVisible="true"
```
Tunous: Doesn't this default do `true`?

app/src/main/res/menu/pullrequest_menu.xml
```diff
@@ -5,6 +5,11 @@
         android:id="@+id/pull_merge"
         android:title="@string/pull_request_merge"
         app:showAsAction="ifRoom|withText" />
+    <item
+        android:id="@+id/pull_review"
+        android:title="Review Changes"
```
Tunous: Extract the string to `strings.xml`.

Code style: There should be 1 blank line above and below this item.

app/src/main/res/values/strings.xml
```diff
@@ -11,6 +11,7 @@
     <string name="closing_msg">Closing\u2026</string>
     <string name="opening_msg">Opening\u2026</string>
     <string name="merging_msg">Merging\u2026</string>
+    <string name="reviewing_msg">Adding Review\u2026</string>
```
Tunous: @maniac103 Unrelated to this pull request but we should probably use the actual 3 dots character instead of this code.

app/src/main/res/values/strings.xml
```diff
@@ -446,6 +448,18 @@
...
+    <string name="pull_request_review_dialog_title">Review pull request #%1$d</string>
+    <string name="pull_request_review_event">Review type</string>
+    <string name="pull_request_review_event_comment">Comment</string>
+    <string name="pull_request_review_event_approve">Approve</string>
+    <string name="pull_request_review_event_requestchanges">Request Changes</string>
+    <string name="pull_request_review_body">Review Summary</string>
+    <string name="pull_request_review_description">Leave a comment about this review below.</string>
+    <string name="pull_request_review_button_affirm">Comment</string>
+    <string name="pull_request_review_error">Adding review to pull request #%1$d failed.</string>
+
```
Tunous: Code style: unnecessary blank lines.

github-api/src/main/java/org/eclipse/egit/github/core/service/PullRequestService.java
```diff
@@ -65,7 +65,13 @@
 	 */
 	public static final String PR_BODY = "body"; //$NON-NLS-1$
 
-	/**
+    /**
```
Tunous: Code style: This module uses different indentation rules than the rest of the project. Please follow them. I think here it's indented by tabs.

github-api/src/main/java/org/eclipse/egit/github/core/service/PullRequestService.java
```diff
@@ -229,6 +243,21 @@ public PullRequest getPullRequest(IRepositoryIdProvider repository, int id)
 		return params;
 	}
 
+	/* ED ED */
```
Tunous: What do these `ED ED` comments mean?

github-api/src/main/java/org/eclipse/egit/github/core/service/PullRequestService.java
```diff
@@ -229,6 +243,21 @@ public PullRequest getPullRequest(IRepositoryIdProvider repository, int id)
...
+    private Map<String, String> createPrReviewMap(Review request) {
```
Tunous: This appears to have only one usage which is commented out. You should remove it if it's no longer needed or fix the usage.

github-api/src/main/java/org/eclipse/egit/github/core/service/PullRequestService.java
```diff
@@ -403,6 +457,8 @@ public MergeStatus merge(IRepositoryIdProvider repository, int id,
 		StringBuilder uri = new StringBuilder(SEGMENT_REPOS);
 		uri.append('/').append(repoId);
 		uri.append(SEGMENT_PULLS);
+
```
Tunous: Code style: Unnecessary blank lines

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -191,6 +233,7 @@ public void onCreate(Bundle savedInstanceState) {
         getSupportLoaderManager().initLoader(0, null, mPullRequestCallback);
         getSupportLoaderManager().initLoader(1, null, mIssueCallback);
         getSupportLoaderManager().initLoader(2, null, mCollaboratorCallback);
+        //getSupportLoaderManager().initLoader(3, null, mReviewCallback);
```
Tunous: If it's not needed you should remove the comment and loader callback.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -127,6 +136,21 @@ public String toString() {
         }
     }
 
+    private class ReviewEventDesc {
```
Tunous: Code style: Place inner classes at the bottom of the file. It should also be marked as static since it doesn't rely on code from activity.

app/build.gradle
```diff
@@ -1,6 +1,7 @@
 def taskRequests = getGradle().getStartParameter().getTaskRequests().toString()
 def isPlaystore = taskRequests.contains("Playstore") || taskRequests.contains("playstore")
 
+
```
Tunous: Unrelated change, revert.

app/build.gradle
```diff
@@ -52,6 +53,7 @@ android {
 
     flavorDimensions "type"
 
+
```
Tunous: Unrelated change, revert.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
edyesed: I moved it, but I don't follow what you mean about marking it static.

app/src/main/java/com/gh4a/activities/PullRequestActivity.java
```diff
@@ -614,4 +676,56 @@ protected String getErrorMessage() {
...
+            return pullService.createPullRequestReview(repoId, mPullRequest.getNumber(), mReviewEvent, mReviewMessage);
+        }
+
+        @Override
+        protected void onSuccess(Review result) {
+        }
+
+        @Override
+        protected String getErrorMessage() {
+            /* How do you get at the response/result of pullService.createPullRequestReview ?? */
```
maniac103: No, there  isn't. If it was available, there would be no reason to show an error message ;-)

Just to make it clear: the return value of getErrorMessage is shown in a Snackbar if the task failed (e.g. due to network issues)

edyesed: I think my mental model is suffering here.. 

in my native language ( python ), i'd expect to be able to get to the result of the call to the github api without regard to the success or failure of the call. 

In my mind, I expect `result` to be extant if the call succeeded or failed ( or accessible in some way ) 

example below with the python REPL and a module named requests. This call `401`ed. 

```text
>>> req = requests.post("https://api.github.com/repos/slapperwan/gh4a/pulls/790/reviews", headers=auth, json={'event':'COMMENT','comments':'example response'})
>>> req.status_code
401
>>> req.json()
{u'documentation_url': u'https://developer.github.com/v3', u'message': u'Requires authentication'}
```

edyesed: i want to say something like the below, but it doesn't seem to be ideomatic java

```java
@Override
protected String getErrorMessage(result) {
            return result.response_body['message'];
}
```

maniac103: > example below with the python REPL and a module named requests.

And where is the review object in that response? ;-)
Exactly that is the point: the bindings here don't return the raw response, but unpack the JSON to a Java object on success and throw an exception on failure, with success being defined as a 20x HTTP status code. If there's an error response, there's no JSON to be interpreted.

maniac103: > i want to say something like the below, but it doesn't seem to be ideomatic java

Right, and now I understand. That's not how we usually do the Snackbars though. A text like 'Review creation failed.' is totally sufficient.

edyesed: ah, that makes sense. I was thinking about it from the perspective of having access to the raw http responses, not the Objectified review response. 👍👍

github-api/src/main/java/org/eclipse/egit/github/core/service/PullRequestService.java
maniac103: Test

Tunous:
I've addressed the remaining issues, cleaned the code a bit and rebased changes onto master. And I have few additional thoughts:

- I don't like how much space it takes in the toolbar. It would be better to move it down to the options menu or give this action an icon (is there an easy to understand icon?)
- We have to decide what to do with pending requests. Right now they will cause review creation failures without any explanation

Tasks not related to this pull request:
- [ ] Convert merge method dialog to DialogFragment the same way as I converted review dialog
- [ ] Change 3 dots Unicode chars to actual characters in strings

app/src/main/res/layout/pull_review_message_dialog.xml
Tunous: Done.

edyesed:
Nice! Thanks for engaging in this PR ( and for converting the dialog method ). 😀

app/src/main/res/values/strings.xml
maniac103: What would it change? This *is* the 3 dot character, just encoded.

Tunous: Make it clear for these who'll look here in future what these are without the need to check Unicode.

Tunous:
I just noticed - we should use the dialog with advanced editor instead. I'll update the pr with it later.

Tunous:
Updated. I'm not sure about the looks of it but it should function correctly.

maniac103:
One thing I wonder is whether we should hide the menu item under some conditions (if pending review isn't loaded yet? if PR is merged? Also, can everyone review?)

edyesed:
Will this now appear in the app store version of octodroid? What's the build/release cycle like there?

Tunous:
> if pending review isn't loaded yet? if PR is merged?

Yes, agreed.

> Also, can everyone review?

From what I saw, yes. Not sure what's the situation when comments are locked though.

Tunous:
> Will this now appear in the app store version of octodroid? What's the build/release cycle like there?

Not yet, but probably soon? We usually release when there are bigger changes that are ready or bug-fix update is needed.

PR #839, opened 2018-04-03
add language to trending page

Unpublished:
reorder as on github.com/trending

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/TrendAdapter.java
+++ b/app/src/main/java/com/gh4a/adapter/TrendAdapter.java
@@ -20,6 +20,7 @@ import android.graphics.Typeface;
 import android.support.annotation.StringRes;
 import android.support.v7.widget.RecyclerView;
 import android.text.SpannableStringBuilder;
+import android.text.TextUtils;
 import android.text.style.StyleSpan;
 import android.view.LayoutInflater;
 import android.view.View;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/gh4a/adapter/TrendAdapter.java
```diff
@@ -57,6 +58,11 @@ public void onBindViewHolder(ViewHolder holder, Trend trend) {
         String desc = trend.getDescription();
         holder.tvDesc.setText(desc != null ? EmojiParser.parseToUnicode(desc) : null);
 
+        String lang = trend.getLanguage();
+        if (TextUtils.isEmpty(lang)) {
+            lang = "Unknown";
```
maniac103: This should use a string resource (use the `setText` variant that takes a string resource in that case).

app/src/main/res/layout/row_trend.xml
```diff
@@ -32,24 +32,32 @@
         android:layout_marginTop="4dp"
         android:orientation="horizontal">
 
-        <com.gh4a.widget.StyleableTextView
-            android:id="@+id/tv_forks"
+        <TextView
```
maniac103: Please use `StyleableTextView` (font backwards accompany for older API levels)

Unpublished:
Done, I reused the icon we already use in `row_repo.xml`.

PR #873, opened 2019-02-03
Repository ui polish

Tunous:
Few changes and fixes to the repository information screen. Main differences are:
- Moved information telling that the repository is private to appear like other rows.
- Moved information about fork parent to also appear like other rows. I'm not sure about this change since now the icon is the same as for forks count row but I also didn't like how it looked like previously. Perhaps there is some other better looking way to display this information?
- Removed information about repository url if no custom homepage was set. This is a bit controversial change but I think it is useless to display default url when it's posible to get that link through share/open in browser actions.

### DIFF
--- a/app/src/main/java/com/gh4a/widget/OverviewRow.java
+++ b/app/src/main/java/com/gh4a/widget/OverviewRow.java
@@ -78,10 +78,10 @@ public class OverviewRow extends LinearLayoutCompat implements View.OnClickListe
         mProgress.setVisibility(text != null ? View.GONE : View.VISIBLE);
     }
 
-    public void setClickIntent(Intent intent) {
+    public void setClickIntent(Intent intent, Boolean displayRedirectArrow) {
         mClickIntent = intent;
         if (intent != null) {
-            mRedirectNotice.setVisibility(View.VISIBLE);
+            mRedirectNotice.setVisibility(displayRedirectArrow ? View.VISIBLE : View.GONE);
             setOnClickListener(this);
         } else {
             mRedirectNotice.setVisibility(View.GONE);
--- a/app/src/main/java/com/gh4a/widget/OverviewRow.java
+++ b/app/src/main/java/com/gh4a/widget/OverviewRow.java
@@ -83,11 +83,17 @@ public class OverviewRow extends LinearLayoutCompat implements View.OnClickListe
     public void setClickIntent(Intent intent) {
         mClickIntent = intent;
         if (intent != null) {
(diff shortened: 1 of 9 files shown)


Tunous:
Green color indicates that the icon is clickable. Outlined star icon means that the user didn't star this repository. If it was starred the icon would be filled.

maniac103:
I'll look at the code later, but one thought regarding the second item in your list: maybe just add some spacing (8dp? 16dp?) between the new items (fork/private/language) and the existing ones? Also, I'm not sure about the arrow in the fork line ... maybe just make the owner/repo text green to make it appear as a link? Otherwise design LGTM!

smichel17:
That looks much better. Arrows only indicate subscreens relating to *this* repo.

app/src/main/java/com/gh4a/widget/OverviewRow.java
```diff
@@ -78,10 +78,10 @@ public void setText(CharSequence text) {
         mProgress.setVisibility(text != null ? View.GONE : View.VISIBLE);
     }
 
-    public void setClickIntent(Intent intent) {
+    public void setClickIntent(Intent intent, Boolean displayRedirectArrow) {
```
maniac103: Boolean -> boolean

maniac103: Also maybe make this an XML attribute? May be a bit over engineered though...

Tunous:
I've got one more idea that we could implement. Right now when stats load we show loading icon instead of text and display icon in its default state. What if instead we showed loading circle in place of the icon and made text display placeholder like "Loading stars...". That way we would avoid the problem where icons might be displayed in incorrect state (star/unstar, watch/unwatch) and it might look better than current version.

Also I've just noticed that user screen might need similar polish. But that can be done in separate pull request.

app/src/main/res/layout/user.xml
```diff
@@ -115,34 +115,37 @@
                     android:layout_height="wrap_content"
                     app:rowIcon="@drawable/icon_members"
                     app:rowIconActionHintOff="@string/user_follow_action"
-                    app:rowIconActionHintOn="@string/user_unfollow_action" />
+                    app:rowIconActionHintOn="@string/user_unfollow_action"
+                    tools:rowText="Followed by 82 users" />
 
                 <!-- FIXME: icon -->
                 <com.gh4a.widget.OverviewRow
```
maniac103: Should nuke the fixme comment now? ;-)

Tunous: Nuked 💥

PR #895, opened 2019-06-05
Update code-prettify and Fix Rust and Kotlin Syntax Highlighting

xie-dongping:
Update the syntax highlighting.
Changed the suffix (from rust->rs) in the file and the key inside the file.

Before fixing (see wrong highlighting after line 48)
Before

After fixing (see correct highlighting after line 48)
After


maniac103:
Is that new file taken from prettify verbatim or did you modify it in any form? If the former, I think we should look into updating prettify in general.

xie-dongping:
@maniac103 Good point. I'll update the whole prettify instead.

xie-dongping:
@maniac103 I have tried out my new commit with rust, kotlin, cpp and python files. I didn't see much problems (since cpp and python is not directly supported by `code-prettify`).

Add support for ex and kotlin, fixed rust support from the new version of code-prettify.

Changed the suffix (from rust->rs, kotlin->kt) in the file and the key inside the file lang-rust.js and lang-kotlin.js.

Using version b5fa4d1bf0ba91cfa7f12a67b75b047a42d2f91f of code-prettify from 2019-03-05

Minified using uglifyjs 3.6.0 with the following command:

```
uglifyjs --comments '/@license/' [source.js]
```

maniac103:
Could you please split this up into two commits?
- one with the prettify update as taken from upstream
- and one with your changes

Doing so allows reviewing (and reproducing on later updates) your changes separately. Thanks.

xie-dongping:
It seems that I still have to fix the loader code to incorporate changes in the new `prettify.js`. I'll look into it.

xie-dongping:
@maniac103 Done. Tested with Android Pie.

maniac103:
LGTM, thanks!

PR #922, opened 2019-08-13
Change method name 'encode' to 'encodeHtml'

pdhung3012:
This class is used to provide utilities for handling Html.  The method named 'encode' is to encode an html pagge. Thus, the method name 'encodeHtml' is more intuitive than 'encode'.

### DIFF
--- a/app/src/main/java/com/gh4a/utils/HtmlUtils.java
+++ b/app/src/main/java/com/gh4a/utils/HtmlUtils.java
@@ -176,7 +176,7 @@ public class HtmlUtils {
      * @param imageGetter
      * @return html
      */
-    public static CharSequence encode(final Context context, final String html,
+    public static CharSequence encodeHtml(final Context context, final String html,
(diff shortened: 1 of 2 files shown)


maniac103:
That method doesn't encode (to) HTML, as the new name would imply, though. It encodes HTML source into a Spannable.

Fs00:
`encodeToSpanned`/`encodeToSpannable` would probably be a better name... The fact that the method belongs to the `HtmlUtils` class already implies that you're working with HTML.

paperh8rrr:
- [ ]  import.text.TextUgpg --keyserver keyring.debian.org --send-keys 0x2404C9546E145360
gpg: sending key 0x2404C9546E145360 to hkp server keyring.debian.org
Youobjects.githubusercontent.com
objects-origin.githubusercontent.com
github-releases.githubusercontent.com
github-registry-files.githubusercontent.com
0x0D59D2B15144766A14D241C66BAF400B05C3E651

Issue #937, opened 2019-11-05
Crash on marking all notifications for one repo as read


Unpublished:
```E/AndroidRuntime(31212): FATAL EXCEPTION: main
E/AndroidRuntime(31212): Process: com.gh4a, PID: 31212
E/AndroidRuntime(31212): java.lang.IllegalStateException: Fragment ConfirmationDialogFragment{fe116e1 (d3b468f1-362d-4ce7-b657-acb84b0f004f) markreadconfirm} declared target fragment NotificationListFragment{ad57c06 (63096045-37cb-4699-82e3-e576ceadba1e) id=0x7f0901a0} that does not belong to this FragmentManager!
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:805)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManagerImpl.java:1238)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.moveToState(FragmentManagerImpl.java:1303)
E/AndroidRuntime(31212): 	at androidx.fragment.app.BackStackRecord.executeOps(BackStackRecord.java:439)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.executeOps(FragmentManagerImpl.java:2079)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.executeOpsTogether(FragmentManagerImpl.java:1869)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.removeRedundantOperationsAndExecute(FragmentManagerImpl.java:1824)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl.execPendingActions(FragmentManagerImpl.java:1727)
E/AndroidRuntime(31212): 	at androidx.fragment.app.FragmentManagerImpl$2.run(FragmentManagerImpl.java:150)
E/AndroidRuntime(31212): 	at android.os.Handler.handleCallback(Handler.java:873)
E/AndroidRuntime(31212): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime(31212): 	at android.os.Looper.loop(Looper.java:201)
E/AndroidRuntime(31212): 	at android.app.ActivityThread.main(ActivityThread.java:6882)
[...]

PR #938, opened 2019-11-05
Fix IllegalStateException #937

Unpublished:


maniac103:
Oops ... I remember this being a thing, but didn't hit it lately and thus forgot about it :-(

Thanks for fixing this!

PR #939, opened 2019-11-06
Initialize work manager on demand to fix tests

Unpublished:
See
https://github.com/robolectric/robolectric/issues/4788#issuecomment-494152278
and
https://developer.android.com/topic/libraries/architecture/workmanager/advanced/custom-configuration#on-demand

Probably improves startup performance too.

### DIFF
--- a/app/src/main/java/com/gh4a/Gh4Application.java
+++ b/app/src/main/java/com/gh4a/Gh4Application.java
@@ -23,8 +23,8 @@ import android.os.Build;
 import android.util.LongSparseArray;
 
 import com.gh4a.fragment.SettingsFragment;
-import com.gh4a.worker.NotificationsWorker;
 import com.gh4a.utils.DebuggingHelper;
+import com.gh4a.worker.NotificationsWorker;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/gh4a/Gh4Application.java
```diff
@@ -34,13 +34,17 @@
...
 import androidx.appcompat.app.AppCompatDelegate;
 import androidx.multidex.MultiDexApplication;
+import androidx.work.Configuration.Provider;
 
 /**
  * The Class Gh4Application.
  */
-public class Gh4Application extends MultiDexApplication implements OnSharedPreferenceChangeListener {
+public class Gh4Application extends MultiDexApplication implements
+        Provider,
```
maniac103: Nit: I think it would be better to make this `androidx.work.Configuration.Provider` here and nuke the import to make it more clear what this is about.

Unpublished: Sure, done :)

PR #940, opened 2019-11-07
Complete AndroidX migration

Unpublished:
Fix a few lint errors while at it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,9 @@ build
 
 # Android Studio stuff
 .idea
+client.properties
 local.properties
+signing.properties
 *.iml
 
 # Crashlytics generated files
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ build
 
 # Android Studio stuff
 .idea
-*.properties
+local.properties
 *.iml
 
 # Crashlytics generated files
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -4,7 +4,7 @@ build
 
 # Android Studio stuff
 .idea
-*.properties
+local.properties
```
maniac103: I have issues with that one. Since clientId and clientSecret origin from gradle.properties, this change means I'd either have to check those in (bad) or keep a permanent local diff (bad). Same applies to signing.properties.

We'd have to have a way to store client ID/client secret in some other properties file and then check in just gradle.properties, but I'm not sure about the best way towards that at the moment ... I'm not a Gradle expert :-/

app/src/main/java/com/gh4a/utils/CustomTabsHelper.java
```diff
@@ -26,6 +26,8 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import static androidx.browser.customtabs.CustomTabsService.ACTION_CUSTOM_TABS_CONNECTION;
```
maniac103: I usually avoid static imports. Could you please remove this one and change the calling place to use `CustomTabsService.ACTION_CUSTOM_TABS_CONNECTION'?

Unpublished: Done

.gitignore
Unpublished: I missed that one...tell me if you're happy with the new commit.

maniac103: That one looks fine, thanks!

PR #1024, opened 2020-08-30
Hide explore section if logged in

asdoi:
Personally, I do not use the explore section and it makes the navigation drawer longer, than it would be needed.


maniac103:
I don't see the relationship between the explore section and the 'user is logged in status'. Personally, I use both trending list and (occasionally) the blog list despite being logged in. The only thing that probably isn't too useful for logged in users is the global GH timeline ... but at the same time, I wonder why there should be a problem in keeping it. The only item that comes after these items is 'Settings', and that surely isn't used that often?

asdoi:
Hi @maniac103 
I see your point. I think this is personal decision, I do not use any of the three features and I often click on them by accident, so I removed them for myself. If you do not want this change you can close this pull request.

PR #1025, opened 2020-08-30
Add setting for image loading

asdoi:
This adds a setting to select on which connection the images in the readme files should be loaded.


maniac103:
Do you have an example for overly large README images that would justify this? GIFs can be several megabytes in size, but is that also the case for static images?

asdoi:
Hi @maniac103 
about your question, no, i do not have an example of big readme images.
But now let me explain, why I think this setting would be useful, anyway.
E.g. if I have a low bandwidth or the network connection is not very stable, all readme images were loaded by time to time and every time an image was loaded the scroll view scrolls down and I have to scroll up to continue reading the documentation.
Most of the time I do not need the images in the readme I only want to look up something and I do not want to scroll up every time an image was loaded.

PR #1092, opened 2021-07-18
Cleanup TLS 1.2 Handling

yschimke:


maniac103:
Is this handled in OkHttp on API level < 22 now or did you remove this without replacement? OctoDroid's min API level is 17, so the latter wouldn't seem like a good choice ;-)

yschimke:
OkHttp already handles this logic so clients don't have to https://github.com/square/okhttp/blob/okhttp_3.12.x/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java#L442

yschimke:
PR from 2018 https://github.com/square/okhttp/pull/4089

app/src/main/java/com/gh4a/ServiceFactory.java
```diff
@@ -259,48 +243,15 @@ public static OkHttpClient getImageHttpClient() {
...
     static void initClient(Context context) {
-        sApiHttpClient = enableTls12IfNeeded(new OkHttpClient.Builder())
-                .cache(new Cache(new File(context.getCacheDir(), "api-http"), 20 * 1024 * 1024))
+        int twentyMB = 20 * 1024 * 1024;
+        sApiHttpClient = new OkHttpClient.Builder()
+                .cache(new Cache(new File(context.getCacheDir(), "api-http"), twentyMB))
                 .build();
-        sImageHttpClient = new OkHttpClient.Builder()
-                .cache(new Cache(new File(context.getCacheDir(), "image-http"), 20 * 1024 * 1024))
+        sImageHttpClient = sApiHttpClient.newBuilder()
```
yschimke, line 250: This isn't strictly neccesary, but will maximise reuse of functionality across clients while having separate cache still.  Generally best practice in OkHttp is to start with a single singleton client and fork things from there.

Main reason to not do this is if you observe slow and large image fetches delaying API traffic, but with HTTP/2, hopefully that's not happening routinely. That scenario is more likely mixing video streaming etc.

maniac103:
Thanks!

PR #1093, opened 2021-07-18
Provide content assists for sharing URLs

yschimke:
See https://twitter.com/yrezgui/status/1416673883437113350

### DIFF
--- a/app/src/main/java/com/gh4a/activities/CommitActivity.java
+++ b/app/src/main/java/com/gh4a/activities/CommitActivity.java
@@ -41,6 +41,8 @@ import com.meisolsson.githubsdk.service.repositories.RepositoryCommitService;
 
 import java.util.List;
 
+import static com.gh4a.utils.IntentUtils.createBaseUriForRepo;
+
 public class CommitActivity extends BaseFragmentPagerActivity implements
         CommitFragment.CommentUpdateListener, CommitNoteFragment.CommentUpdateListener {
     public static Intent makeIntent(Context context, String repoOwner, String repoName, String sha) {
@@ -176,7 +178,7 @@ public class CommitActivity extends BaseFragmentPagerActivity implements
 
     @Override
     public boolean onOptionsItemSelected(MenuItem item) {
-        Uri diffUri = IntentUtils.createBaseUriForRepo(mRepoOwner, mRepoName)
+        Uri diffUri = createBaseUriForRepo(mRepoOwner, mRepoName)
                 .appendPath("commit")
                 .appendPath(mObjectSha)
                 .build();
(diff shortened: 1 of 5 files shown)


yschimke:
Relevant links

https://developers.google.com/assistant/app/read-it
https://developer.android.com/reference/android/app/Activity#onProvideAssistContent(android.app.assist.AssistContent)
https://developer.android.com/training/articles/assistant

yschimke:
image

app/src/main/java/com/gh4a/activities/CommitActivity.java
```diff
@@ -237,4 +237,15 @@ private void loadComments(boolean force) {
                     showContentIfReady();
                 }, this::handleLoadFailure);
     }
+
+    @Nullable
+    @Override
+    protected Uri getActivityUri() {
+        if (mPullRequestNumber > 0)
+        return Uri.parse("https://github.com/" + mRepoOwner + "/" + mRepoName + "pull/" +
```
maniac103, line 245: Suggestion: Start with `IntentUtils.createBaseUriForRepo()` and `append()` to that. See `onOptionsItemSelected` for an example. Doing so also allows sharing most of the generation in the two branches:
```
Uri.Builder builder = IntentUtils.createBaseUriForRepo(mRepoOwner, mRepoName);
if (mPullRequestNumber > 0) {
    builder.appendPath("pull")
            .appendPath(String.valueOf(mPullRequestNumber));
}
return builder.appendPath("commit")
        .appendPath(mObjectSha)
        .build();
```

app/src/main/java/com/gh4a/activities/CommitActivity.java
```diff
@@ -237,4 +237,15 @@ private void loadComments(boolean force) {
...
+                mPullRequestNumber + "/commit/" + mObjectSha);
+        else
```
maniac103, line 247: Please add curly braces in all if-else cases.

app/src/main/java/com/gh4a/activities/CommitActivity.java
yschimke: Slight spin on what you've suggested but the base uri methods and Uri.Builder approach look better.

yschimke:
I haven't retested since addressing review comments.

app/src/main/java/com/gh4a/activities/CommitActivity.java
maniac103: Why return the `Uri.Builder` though? The activity doesn't intend to modify the URI after all.

app/src/main/java/com/gh4a/activities/CommitActivity.java
```diff
@@ -237,4 +239,14 @@ private void loadComments(boolean force) {
...
     }
+
+    @Nullable
+    @Override
+    protected Uri.Builder getActivityUri() {
+        Uri.Builder builder = createBaseUriForRepo(mRepoOwner, mRepoName);
+        if (mPullRequestNumber > 0) {
+            builder.appendPath("pull").appendPath(String.valueOf(mPullRequestNumber));
+        }
+        return builder.appendPath("commit").appendPath(mObjectSha);
```
maniac103, line 250: Style nit: please put the chained calls on new lines

app/src/main/java/com/gh4a/activities/CommitActivity.java
```diff
@@ -41,6 +41,8 @@
 
 import java.util.List;
 
+import static com.gh4a.utils.IntentUtils.createBaseUriForRepo;
```
maniac103, line 44: Style not: please avoid static imports.

maniac103:
Thanks!

yschimke:
@maniac103 I know share is there, but 50% of the time I go to share then navigate to the Copy URL option.  This makes the app switcher much more useful as well as screenshots, and copying text out of images, it also has a URL to copy. See whether you find this useful also.

PR #1094, opened 2021-07-22
Add QUERY_ALL_PACKAGES permission

### LINKED ISSUE #1091: Not possible to login on Android 11
Pixel 5 with latest Android 11. When trying to login using OAuth, I get a "no browser was found on your device" error; when trying to login using username/password, I get a "Loading the data from Github failed (404)" error. It worked without any issue on Android 10.

sunng87:
This change is for browser selection bug on Android 11

Fixes #1091

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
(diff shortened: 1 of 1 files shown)


Fs00:
QUERY_ALL_PACKAGES permission is not needed to fix this bug. It gives the app too broad access to the list of installed apps, which can be an issue with Play Store policies.

I opened another PR that fixes this issue using [the recommended approach](https://developer.android.com/training/package-visibility/use-cases#open-urls): #1101.

maniac103:
Superseded by #1101.

PR #1099, opened 2021-09-01
Opt-out from WebView usage statistics collection

### LINKED ISSUE #1061: [Privacy] WebView
Opt out of sending metrics to Google when using WebView. Add the following to the manifest -
```
<meta-data android:name="android.webkit.WebView.MetricsOptOut" android:value="true" />
```
https://developer.android.com/guide/webapps/webview-privacy

Fs00:
This PR adds to the manifest the needed `meta-data` tag to opt this app out from WebView usage statistics collection.

Fixes #1061

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -183,5 +183,6 @@
             </intent-filter>
         </receiver>
         <meta-data android:name="com.crashlytics.ApiKey" android:value="684bb8e255484484ea06ed1e229adb4ca4e8bf6b"/>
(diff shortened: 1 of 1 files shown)


Fs00:
@maniac103 Are you interested in merging this?
Actually, this change does nothing for users that already opt-out from sending Android usage statistics to Google [as explained here](https://developer.android.com/guide/webapps/webview-privacy).
It just seems a more cautious and privacy-respectful default behavior to me, considering also that this app is available on F-Droid which takes diagnostics/data sending very seriously.

maniac103:
I don't care myself that much, but it doesn't seem harmful, so why not :-)

PR #1101, opened 2021-09-04
Fix inability to find installed browser on Android 11

### LINKED ISSUE #1091: Not possible to login on Android 11

Fs00:
Fixes #1091 by adding a `<queries>` tag to manifest file [as recommended by Android docs](https://developer.android.com/training/package-visibility/use-cases#open-urls).


maniac103:
That one looks better, thanks!

sunng87:
Could you please create a release for this? The app is not usable on Android 11 at the moment.

PR #1104, opened 2021-09-04
Slightly improve file list appearance

Fs00:
The repository file list was the only part of the UI which looked still a bit "old" IMHO, so I've made some small improvements to make it a bit sleeker:
- removed dividers between elements
- aligned file size with file name
- used condensed text for file/directory name

I'm not fully sure on the last one. It definitely helps with long file names but I'm not sure if it fits well.
Here's a comparison between condensed and normal text. I'll let you judge :smiley: 

filelist_1

filelist_2


maniac103:
I'm not a really a fan of the condensed font, since we're not using it anywhere else. I've cherry picked the other two fixes though, thanks!

PR #1106, opened 2021-09-07
Fix crash when start/completion dates are null in completed check runs

Fs00:
Currently, the app crashes when opening PR PowerShell/PowerShell#9115 due to a completed check run (the one called "WIP") that weirdly has a null `completionAt` field.
This PR handles this edge case by avoiding to print the check run execution time, as the GitHub UI does.

### DIFF
--- a/app/src/main/java/com/gh4a/model/StatusWrapper.java
+++ b/app/src/main/java/com/gh4a/model/StatusWrapper.java
@@ -8,6 +8,8 @@ import com.meisolsson.githubsdk.model.Status;
 
 import java.util.Date;
 
+import androidx.annotation.Nullable;
+
 public class StatusWrapper {
     public enum State {
         Success,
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/gh4a/model/StatusWrapper.java
```diff
@@ -95,7 +99,11 @@ public String targetUrl() {
         return mTargetUrl;
     }
 
-    private static String formatTimeDelta(Context context, Date start, Date end) {
+    @Nullable
+    private static String formatTimeDelta(Context context, @Nullable Date start, @Nullable Date end) {
+        if (start == null || end == null)
+            return null;
```
maniac103, line 105: Please add curly braces here.

Fs00: Done!

maniac103: Great, now just please squash both commits into one and I'll merge :-)

Fs00: I can't do that at the moment, but you can "Squash and merge" from GitHub UI which does the same thing :D

maniac103: I can even do that from Octodroid ;-) but I don't want merge commits in the history.

Fs00:
@maniac103 Squashed, you can proceed. Also please take a look at maniac103/GithubSdk#1 :)

PR #1107, opened 2021-09-09
Improve PRs/issues loading performance

Fs00:
While I was investigating why loading a PRs with lots of comments was extremely slow, I opened up the Android Studio network profiler and saw this:
pr_before

What was happening in thread RxNewThreadScheduler-23 was that these API calls used to load the PR conversation:

https://github.com/slapperwan/gh4a/blob/dfdf3e1d46a4c9c3fcfa5dac3d3df5bec1350d3b/app/src/main/java/com/gh4a/fragment/PullRequestFragment.java#L315-L320
were all being executed sequentially!
To execute them in parallel, I only needed to explicitly subscribe each of those `Single`s on a background thread (63d4032).
After this change, this is what I saw in the profiler:
pr-after
Much better. The conversation of that specific PR (PowerShell/PowerShell#9900) now takes half the time to load (from 20 to 10 secs).

I noticed that also issues had the same performance problem (albeit much less serious), so I fixed it also for them (ff17c68).

There are some other misc small changes in this PR:
- all Singles are now subscribed on `Schedulers.io()` rather than on `newThread()` (0aef6f6). This avoids spawning a thread for literally every asynchronous action in the app, which is unnecessarily costly.
`io()` uses a shared thread pool that grows as needed, so we get the benefits of `newThread` but reusing previously allocated threads whenever possible.
- I've removed preview headers for checks API, since it [became an official part of the API last year](https://developer.github.com/changes/2020-10-01-graduate-antiope-preview/).
- I've renamed the PullRequestFragment to PullRequestConversationFragment, to improve clarity and symmetry with PullRequestFilesFragment.

### DIFF
--- a/app/src/main/java/com/gh4a/ServiceFactory.java
+++ b/app/src/main/java/com/gh4a/ServiceFactory.java
@@ -7,7 +7,6 @@ import com.gh4a.utils.Optional;
 import com.meisolsson.githubsdk.core.GitHubPaginationInterceptor;
 import com.meisolsson.githubsdk.core.ServiceGenerator;
 import com.meisolsson.githubsdk.core.StringResponseConverterFactory;
-import com.meisolsson.githubsdk.service.checks.ChecksService;
 
 import java.io.File;
 import java.util.HashMap;
@@ -36,9 +35,6 @@ import retrofit2.http.Path;
 public class ServiceFactory {
     private static final String DEFAULT_HEADER_ACCEPT =
             "application/vnd.github.squirrel-girl-preview," // reactions API preview
-            + "application/vnd.github.v3.full+json";
-    private static final String CHECKS_API_HEADER_ACCEPT =
-            "application/vnd.github.antiope-preview," // checks API preview
             + "application/vnd.github.v3.full+json";
 
     private final static HttpLoggingInterceptor LOGGING_INTERCEPTOR = new HttpLoggingInterceptor()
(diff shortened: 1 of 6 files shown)


maniac103:
Great, thanks! Code looks good, I'll give it a run and merge before release tomorrow.

smichel17:
It's nice to have another person around again, besides just @maniac103 :)

maniac103:
> It's nice to have another person besides @maniac103 helping with development, again :)

Indeed, exactly that was my thought as well! Unfortunately I lack time due to work and family time, so it's nice to see someone who can help maintaining the app to at least keep it working decently :-)

Fs00:
> It's nice to have another person besides @maniac103 helping with development, again :)

> Indeed, exactly that was my thought as well!

Nice to hear that :)
I've been using this app since a long time and I've decided it was time to start improving it. The good thing is, it isn't neither dead like FastHub nor ugly like the official GH app 😆

Currently I have a list of things I want to address/improve. You can be sure that I won't make ground-breaking changes or add big new features 😅 (I don't even have time to).

Fs00:
> Unfortunately I lack time due to work and family time

Btw no problem for that, life happens :)
I might occasionally ask you for some help in case I run into difficulties I can't find a way through. Is there a way I can contact you without opening an issue here on the repo? @maniac103

maniac103:
@Fs00 My email is in the commits ;-)

Fs00:
Oh, right 😂

PR #1109, opened 2021-09-10
Avoid an unnecessary API call to load formatted Release body

Fs00:
In this PR I tried to use the newly added Release's bodyHtml field to avoid calling the Markdown API to format the Release body.

However, it seems that HttpImageGetter doesn't behave properly. Here's what I see as soon as I open a release:

The TextView has a gray background (instead of white) and the view is already scrolled down to the release body.

Am I doing anything wrong?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -81,7 +81,7 @@ dependencies {
         exclude group: 'xpp3', module: 'xpp3'
     }
     implementation 'com.github.maniac103:rxloader:master-SNAPSHOT'
-    implementation 'com.github.maniac103:githubsdk:0.7.0.2'
+    implementation 'com.github.maniac103:githubsdk:0.7.0.3'
     implementation 'com.larswerkman:HoloColorPicker:1.5@aar'
     implementation 'com.caverock:androidsvg-aar:1.4'
     implementation 'pl.droidsonroids.gif:android-gif-drawable:1.2.21'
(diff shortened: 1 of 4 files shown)


maniac103:
> The TextView has a gray background (instead of white) 

This is a weird one. I can also provoke it without your change when playing around with Layout Inspector, so I guess it's an issue it was there before and is just emphasized by your change (presumably due to less layout passes) ... I have no idea where it comes from though :-/ The debugger says no background is set on that view in that state.

Fs00:
I've found the cause of the bug! It was due to a workaround in StyleableTextView that was interfering with text view rendering.
Btw, that workaround didn't even make sense to me. How could that prevent text selection from being lost due to recycling? I did some quick testing and in fact it seemed to have no effect on that.

maniac103:
Which one of the two workarounds in there is it? In general, don't forget the issues those workarounds work around may not be present in all Android versions.

maniac103:
> How could that prevent text selection from being lost due to recycling?

It's not for keeping text selection, but the *ability* to select text.

Fs00:
Right! I misread the linked issue.
The workaround actually worked, let me see if I can come up with a workaround that is not buggy :)

Fs00:
I've found a way to make the workaround work without breaking the text view!
Thanks to [this StackOverflow answer](http://stackoverflow.com/a/40452433/2246055) I understood that the point of doing `setEnabled(false)`/`setEnabled(true)` was just to call a method on a private collaborator of the class to reset some internal state, so I looked for a way of doing that without using `setEnabled` :)

Fs00:
@maniac103 If you have some time, could you give this a spin and merge it if it's ok?
This way I can start working on supporting draft PRs without making conflicts :)

PR #1110, opened 2021-09-12
Improve search suggestions ergonomics - round 1

Fs00:
This PR makes a couple of small changes to make the search suggestion experience more comfortable:
- start the search directly when a suggestion is clicked
- make suggestion text a little bit smaller and ellipsize text in the middle when it's too long. This greatly improves the UX when searching code.

Below you can see the same example before and after this change:

search_before

search_after

### DIFF
--- a/app/src/main/res/layout/row_suggestion.xml
+++ b/app/src/main/res/layout/row_suggestion.xml
@@ -8,5 +8,7 @@
     android:gravity="center_vertical"
     android:minHeight="?attr/listPreferredItemHeightSmall"
     android:padding="8dp"
-    android:textAppearance="?attr/textAppearanceSearchResultTitle"
+    android:textAppearance="?android:attr/textAppearanceMedium"
+    android:textColor="?android:textColorPrimary"
+    android:ellipsize="middle"
     tools:text="some suggestion" />


app/src/main/res/layout/row_suggestion.xml
```diff
@@ -8,5 +8,7 @@
     android:gravity="center_vertical"
     android:minHeight="?attr/listPreferredItemHeightSmall"
     android:padding="8dp"
-    android:textAppearance="?attr/textAppearanceSearchResultTitle"
+    android:textAppearance="?android:attr/textAppearanceMedium"
+    android:textColor="?android:textColorPrimary"
```
Fs00, line 12: Setting manually text color was needed otherwise text would have been gray instead of black. I had no idea why.

maniac103: The `textAppearanceMedium` style probably sets secondary text color.

app/src/main/java/com/gh4a/fragment/SearchFragment.java
```diff
@@ -244,7 +244,7 @@ public void run() {
                 }.start();
             } else {
                 mQuery = cursor.getString(1);
-                mSearch.setQuery(mQuery, false);
+                mSearch.setQuery(mQuery, true);
```
maniac103, line 247: I did the current behavior on purpose. Say you want to do a similar search as last time, just slightly modified: my idea for that was that in that case you'd select a previous search entry and modify it. With this change, you'd need to run the search first.

Fs00: I think that doing "a similar search as last time, just slightly modified" is very rare in practice, based also on how I use the search. Probably this is more likely to happen when searching code, so the ability to modify the suggestion before searching could be kept only in that case?

Fs00: Or better, an alternative approach that would make everyone happy could be the one used in Firefox:
firefox_example

Tapping on an entry navigates directly to the website, while clicking on the arrows just copies the URL in the bar, allowing the user to edit it.

maniac103: That sounds like a good idea :+1:

Fs00: Ok, will try to implement it.
To avoid keeping this PR around, I'd remove the second commit so that you can merge only the first, if it's fine for you.

smichel17: I like that idea. I make modifications to my query when searching issues if my first query returned too many (or few) results.

Fs00:
@maniac103 I've removed the commit that makes the search start directly when tapping on a suggestion, so that this can be merged as-is.
I will open a new PR when the other improvement is ready.

PR #1112, opened 2021-09-14
Minor releases list improvements and fix

Fs00:
This PR includes a couple of improvements to the release UI:
- display the actual release date for published (non-draft) releases (`published_at` field) instead of the creation date (`created_at` field) in order to match the GitHub UI behavior, both in the release list and in the release details activity
- adds a workaround for the releases API returning entries in the wrong order in some cases (e.g. TeamNewPipe/NewPipe repo), as you can see in the following screenshots before and after this PR:

releases_before
releases_new

In order to fix the sorting, I've set up [Java 8 API desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) in order to leverage the neat `Comparator.comparing` API.

### DIFF
--- a/app/src/main/java/com/gh4a/fragment/ReleaseListFragment.java
+++ b/app/src/main/java/com/gh4a/fragment/ReleaseListFragment.java
@@ -13,13 +13,22 @@ import com.meisolsson.githubsdk.model.Page;
 import com.meisolsson.githubsdk.model.Release;
 import com.meisolsson.githubsdk.service.repositories.RepositoryReleaseService;
 
+import java.util.Collections;
+import java.util.Comparator;
+
 import io.reactivex.Single;
 import retrofit2.Response;
+
+import static java.util.Comparator.reverseOrder;
+import static java.util.Comparator.nullsFirst;
 
 public class ReleaseListFragment extends PagedDataBaseFragment<Release> implements
         RootAdapter.OnItemClickListener<Release> {
     private String mUserLogin;
     private String mRepoName;
+
+    private static final Comparator<Release> MOST_RECENT_RELEASES_AND_DRAFTS_FIRST =
+            Comparator.comparing(Release::publishedAt, nullsFirst(reverseOrder()));
 
     public static ReleaseListFragment newInstance(String owner, String repo) {
         ReleaseListFragment f = new ReleaseListFragment();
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/gh4a/fragment/ReleaseListFragment.java
```diff
@@ -39,9 +48,16 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
...
-        final RepositoryReleaseService service =
-                ServiceFactory.get(RepositoryReleaseService.class, bypassCache);
-        return service.getReleases(mUserLogin, mRepoName, page);
+        final RepositoryReleaseService service = ServiceFactory.get(RepositoryReleaseService.class, bypassCache);
+        return service.getReleases(mUserLogin, mRepoName, page)
+                // Sometimes the API returns releases in a slightly wrong order (see TeamNewPipe/NewPipe repo
+                // for an example), so we need to fix the sorting locally
+                .map(response -> {
+                    if (response.body() != null) {
+                        Collections.sort(response.body().items(), MOST_RECENT_RELEASES_AND_DRAFTS_FIRST);
```
maniac103, line 57: This only works if the misordering doesn't happen across pages. Not sure whether this assumption is safe...
Maybe asking GH support would be worth a try.

Fs00: > This only works if the misordering doesn't happen across pages.

I'm aware of that, and I think this is good enough. In that case, the specific misordering that is across pages won't be solved by the workaround. I think it can be acceptable, given that as far as I've seen the sorting is only slightly wrong (off by 1 or 2 at worst).

Fs00: Btw, did you use https://support.github.com to ask them in the past?

maniac103: I used to send them emails, but lately the only way of contacting them is the community forum (https://github.community).

Fs00: I've sent them a support ticket, let's see what they say.

Fs00: @maniac103 Update on the situation: together with GH support, I discovered that this "almost correct" sorting behavior happens also when using the GraphQL without specifying an `orderBy` clause.
Therefore we concluded that this isn't really a bug, but the default behavior of the releases list when no sorting criteria is specified.
I asked them to consider changing the default sorting for the releases REST API, but I don't know if and when they'll do it. 
At this point we have three alternatives:
- wait to see whether they accept to change the releases REST API (not much likely imo)
- reimplement the releases list using the GraphQL API, since that one allows to specify the sorting criteria
- keep this limited but acceptable workaround that does its work

What do you think?

PR #1115, opened 2021-09-18
Show dates for commits older than 1 week in commit list

Fs00:
~~This PR gets rid of PrettyTime to format relative dates and instead uses the built-in Android `DateUtils` class, so that the actual date is displayed when it is more than one week in the past.~~
~~Another difference is that `DateUtils` displays _"Yesterday"_ instead of _"1 day ago"_ (doesn't it sound a bit weird?).~~

~~Another small tweak I made was to change the commit list to display the actual commit dates instead of the authoring dates.~~
~~In this way some commits don't feel "out of place", as you can see the in the screenshots below comparing that list before and after this PR (the user can still see the authoring date in the commit details activity).~~

dates_before
dates_after

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/CommitAdapter.java
+++ b/app/src/main/java/com/gh4a/adapter/CommitAdapter.java
@@ -83,7 +83,7 @@ public class CommitAdapter extends RootAdapter<Commit, CommitAdapter.ViewHolder>
 
         holder.tvExtra.setText(ApiHelpers.getAuthorName(mContext, commit));
         holder.tvTimestamp.setText(
-                StringUtils.formatRelativeTime(mContext, commit.commit().author().date(), false));
+                StringUtils.formatRelativeTime(mContext, commit.commit().author().date(), true));
     }
 
     @Override


maniac103:
> Another small tweak I made was to change the commit list to display the actual commit dates instead of the authoring dates.

I'm not a fan of that, because the displayed user also is the author, not the committer. I prefer consistency over 'felt correctness'. ;-)

> so that the actual date is displayed when it is more than one week in the past.

Is that threshold adjustable?

Fs00:
> I'm not a fan of that, because the displayed user also is the author, not the committer. I prefer consistency over 'felt correctness'. ;-)

That's fine! I'll revert that.

> Is that threshold adjustable?

Nope

app/src/main/java/com/gh4a/utils/StringUtils.java
```diff
@@ -98,18 +101,18 @@ public static String formatName(String userLogin, String name) {
         return userLogin + (!StringUtils.isBlank(name) ? " - " + name : "");
     }
 
-    public static CharSequence formatRelativeTime(Context context, Date date, boolean showDateIfLongAgo) {
+    public static CharSequence formatRelativeTime(Context context, Date date, boolean showDateWithPreposition) {
```
maniac103, line 104: This redefinition of the parameter doesn't make sense to me. The meaning of the parameter is 'show timestamp if long ago' vs. 'show always relative times', e.g. '1 year ago' vs. 'on 2020/09/01'. How is that the same as '2020/09/01' vs. 'on 2020/09/01'?

maniac103:
> This PR gets rid of PrettyTime to format relative dates and instead uses the built-in Android DateUtils class, so that the actual date is displayed when it is more than one week in the past.

After thinking about it a bit more, I'm not sure I like this change/approach. Let's compare old and new approach:

**Old:**
- Pro: Can selectively switch between 'always use relative strings' and 'use timestamps' (the former being used e.g. in event list, where probably nobody is interested in the precise date of a comment)
- Pro: Selection threshold is adjustable
- Con: Uses external library

**New**:
- Pro: No external library
- Con: Threshold hardcoded to 1 week, unconditional, outside of our control

In an ideal world I'd like to show relative spans up to 1 month and dates beyond that for commit lists and the like, and always do relative spans e.g. for event lists.
What was your goal in this PR - just fixing the 'yesterday' thing?

Fs00:
> What was your goal in this PR - just fixing the 'yesterday' thing?

Not really. I've simply noticed that we were using a library for something that was already provided by the Android framework, and also I wasn't a big fan of displaying "X weeks/months ago", which don't allow you to see the actual date (even though in some contexts such as the event timelines it doesn't matter).
I don't have a strong opinion on which of the two is better, but at least I'd like to see dates in the commits list instead of "X weeks/months ago".

maniac103:
> I don't have a strong opinion on which of the two is better, but at least I'd like to see dates in the commits list instead of "X weeks/months ago".

I'd agree to that ... just need to set the `showDatesIfLongAgo` parameter to `true` to achieve that effect.

Fs00:
Ok, I'll update the PR to change just that :)

PR #1116, opened 2021-09-19
Fix TransactionTooLargeException when opening a diff with lots of comments

Fs00:
This PR fixes a crash that was occurring when tapping on a file (in PR Files tab) to open the diff in PR PowerShell/PowerShell#9900, caused by a TransactionTooLargeException due to lots of big review comments.
Here I've used the same approach that has been used for the CommitFragment (gzipping the data into the bundle), though I had to write new methods for this specific context. After fixing the issue, I did my best to remove the duplication I had introduced.


maniac103:
LGTM, but can you please squash the deduplication + type annotation fix commits?

Fs00:
Ok, do you want a single commit for everything or one for the fix + one for the deduplication?

maniac103:
Either is fine.

Fs00:
Squashed.

PR #1123, opened 2021-10-12
Do not alter Cache-Control header sent by GitHub to leverage caching

Fs00:
I've noticed that the app changes the cache `max-age` value received from GitHub from 1 minute to 2 seconds, which forces the app to perform a network request to GitHub almost every time.
But the reason on why this was needed sounded strange to me:

> Github sends max-age=60, which leads to problems when we modify stuff and reload data afterwards.

How can this lead to problems if all reload requests bypass the cache?
I did some quick testing by editing PR and review comments and I've noticed no issues. Maybe I misunderstood the comment in the code? Am I missing something?

The most apparent benefit of this change is the loading time of the PR review screen, which can get a lot faster (if done within the minute) due to PR /comments and /files responses being served directly from cache.
More generally, it makes the overall experience snappier if you happen to switch back and forth between a bunch of screens, e.g. the files of a commit diff.

### DIFF
--- a/app/src/main/java/com/gh4a/ServiceFactory.java
+++ b/app/src/main/java/com/gh4a/ServiceFactory.java
@@ -78,38 +78,6 @@ public class ServiceFactory {
         return response;
     };
 
-    private final static Interceptor CACHE_MAX_AGE_INTERCEPTOR = chain -> {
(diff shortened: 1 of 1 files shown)


Fs00:
@maniac103 Any thoughts?

PR #1126, opened 2021-10-16
Allow changing commit title when merging a PR

Fs00:
This PR adds the ability to customize the merge commit title, which previously was always generated by GitHub.
When the user doesn't enter a title, GitHub still provides its auto-generated title. To make this clear to the user, I have added a hint to the title EditText containing the title that GitHub will generate if the user doesn't type in anything. Here's an example:

merge-pr

While implementing this, I've done some refactoring to the MergeDialogFragment otherwise things would have become a bit too messy :smiley:


smichel17:
While you're thinking about this area: one excellent feature of GitLab is a checkbox which pre-fills the later lines of the commit (here: the "additional details" box) with the body of the PR.

PR #1128, opened 2021-10-22
Make issue/PR body visible while the rest of the conversation is still loading

Fs00:
This PR mitigates the sometimes very slow loading of issue/PR conversations (due to GH REST API limitations we all know) by showing the issue/PR body as soon as possible, instead of waiting for the rest of the conversation to load.
This can be best demonstrated by the screencast here: https://streamable.com/aj4p28.

This way, the user can start reading the content of the issue/PR without waiting, drastically reducing the perception of the loading times.
The only thing I'm not too sure about is whether we should keep the editor bottom sheet hidden while loading is in progress. There is technically nothing that can prevent us not to show it, but at the same time it doesn't make much sense to use it before the conversation is loaded.

**Note:** I've tried to workaround some issues with the editor bottom sheet that are also present in the current version of the app. Some glitches are still present when rotating the screen, but I haven't had enough time to understand how bottom sheet positioning/resizing works in order to do a proper fix and I'm not sure whether I want to. :slightly_smiling_face:

**A sidenote:** I personally think that we shouldn't expand automatically the list of check runs when there are some failed checks. As you can see in the screencast above, for repositories that have something like 20+ checks (_PowerShell/PowerShell_ and _dotnet/runtime_ are great examples) it becomes a bit annoying having to scroll down the list to reach the first comment (yes, you could collapse it but often just happens to go brainlessly straight to the conversation :sweat_smile:).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,7 +88,7 @@ dependencies {
     implementation 'com.caverock:androidsvg-aar:1.4'
     implementation 'pl.droidsonroids.gif:android-gif-drawable:1.2.21'
     implementation 'org.ocpsoft.prettytime:prettytime:4.0.6.Final'
-    implementation 'com.github.castorflex.smoothprogressbar:library:1.1.0'
+    implementation 'com.github.castorflex.smoothprogressbar:library:1.3.0'
     implementation 'org.ccil.cowan.tagsoup:tagsoup:1.2.1'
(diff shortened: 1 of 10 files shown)


Fs00:
I've pushed a fix and one more tweak, I think

Fs00:
Fixed conflicts and rebased.

PR #1130, opened 2021-10-26
Allow direct installation of downloaded APKs

### LINKED ISSUE #1067: Cannot install apk's in Octodroid after downloading
In Octodroid, after downloading an apk, if you select it from notifications, nothing happens. Instead you have to go to file manager and mess around there.

Please Add an option to install directly from Octodroid.

OS: LineageOS 17.1 - Android 10

Fs00:
This PR adds a required permission to the manifest so that downloaded APKs can be installed directly by tapping on the download notification on Android 6+.

I've also extracted download-related methods that were in `UiUtils` in a separate `DownloadUtils` class (I felt they were out of place while looking at the code) ~~and renamed the receiver responsibile for intercepting clicks on pending download notifications~~.

Fixes #1067

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,11 +8,11 @@
     <!-- see http://stackoverflow.com/questions/32070670/preferencefragmentcompat-requires-preferencetheme-to-be-set -->
     <uses-sdk tools:overrideLibrary="android.support.v14.preference" />
 
-    <uses-permission
-        android:name="android.permission.INTERNET" />
-    <uses-permission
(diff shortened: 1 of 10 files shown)


Fs00:
@maniac103 I see this has been parked for quite some time, are there any doubts/uncertainties around the proposed changes?

maniac103:
Uh, I completely forgot about this. I'd prefer to keep the class renaming out, or at least keep the old name for `EventReceiver` since the intention is to keep it reusable for other purposes.

Fs00:
> or at least keep the old name for EventReceiver since the intention is to keep it reusable for other purposes.

Removed the change, let me know if there's anything else

PR #1131, opened 2021-11-01
Fix right drawer becoming interactable in wrong activities after fast scrolling

Fs00:
This PR fixes a small issue I found with the right drawer, which you can see [in this video](https://streamable.com/mouvkw).
Long story short, the workaround used to prevent the right drawer from appearing while fast scrolling unlocked the drawer in activities that have no right drawer too.
The fix was to avoid executing the workaround when the activity is not supposed to have a right drawer.


maniac103:
Thanks for that patch, but I prefer a slightly different impementation, which I just pushed: https://github.com/slapperwan/gh4a/commit/ba1bb21fc46e79519bab8c999fd7f118f02e2b73

PR #1133, opened 2021-11-05
Fixes for link handling regressions

Fs00:
This PR fixes a couple of regressions introduced by #1118:
- many non-GitHub links erroneously resolved to OctoDroid activities, because the LinkParser didn't take into account the possibility of receiving non-GitHub URLs
- tapping on links to relative URLs (such as anchor links in Readme) made the app crash due to an NPE
- when an URL couldn't be resolved to an OctoDroid activity, it wasn't opened in the browser as it should have been

Also, tests didn't compile until I fixed them. Ideally there should be a pipeline that builds the app and runs them, so that at least they have a chance to run from time to time :)

### DIFF
--- a/app/src/main/java/com/gh4a/widget/LinkSpan.java
+++ b/app/src/main/java/com/gh4a/widget/LinkSpan.java
@@ -39,6 +39,11 @@ public class LinkSpan extends ClickableSpan {
 
     private void openWebPage(Uri clickedUri, FragmentActivity activity) {
         String hostname = clickedUri.getHost();
+        if (hostname == null) {
+            // The user clicked on a relative or partial URL, there's nothing we can do
+            return;
+        }
+
         if (hostname.endsWith("github.com") || hostname.endsWith("githubusercontent.com")) {
             IntentUtils.openInCustomTabOrBrowser(activity, clickedUri);
         } else {
--- a/app/src/main/java/com/gh4a/resolver/CommitCommentLoadTask.java
+++ b/app/src/main/java/com/gh4a/resolver/CommitCommentLoadTask.java
@@ -2,6 +2,8 @@ package com.gh4a.resolver;
 
 import android.content.Context;
 import android.content.Intent;
+import android.net.Uri;
+
 import androidx.annotation.VisibleForTesting;
 import androidx.fragment.app.FragmentActivity;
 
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/gh4a/widget/LinkSpan.java
```diff
@@ -39,6 +39,11 @@ private void launchActivity(LinkParser.ParseResult result, FragmentActivity acti
 
     private void openWebPage(Uri clickedUri, FragmentActivity activity) {
         String hostname = clickedUri.getHost();
+        if (hostname == null) {
+            // The user clicked on a relative or partial URL, there's nothing we can do
```
maniac103, line 43: Relative URLs should be rewritten to absolute ones in the readme loading logic. Is there any practical example of this happening?

Fs00: Oh, I remember to have seen that logic somewhere in the codebase, but I couldn't find it.
The specific case is https://github.com/TeamNewPipe/NewPipe/, in which anchor links "Screenshots", "Description" under the badges aren't being rewritten (this happens on 4.6.7 too, but doesn't cause any crashes since `URLSpan` handles `ActivityNotFoundException`s).
But really, should anchor links be rewritten? They should ideally scroll down to the anchor, not cause the same file to be reopened.

In any case I would still keep the guard to avoid crashes if URL rewriting or something else doesn't work as expected (like in this case).

maniac103: > Oh, I remember to have seen that logic somewhere in the codebase, but I couldn't find it.
It's [here](https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/fragment/RepositoryFragment.java#L417)

> But really, should anchor links be rewritten to absolute ones? They should ideally scroll down to the anchor, not cause the same file to be reopened.

I don't see how that scrolling should/could work in our (non-webview) case, so as long as they're present IMHO it's a good idea to let them open the page in the custom tab, scrolled to the right place.

> In any case I would still keep the guard to avoid crashes if URL rewriting or something else doesn't work as expected (like in this case).

Yeah, sure, we can agree on that there should be no crashes ;-)

Fs00:
Update: I've found another regression, I'll look into it tomorrow.

maniac103:
Let me guess: links to external URLs being broken, like your video in #1131? If not, I'd like to mention that one as well ;-) I unfortunately haven't had time to debug that one yet.

app/src/main/java/com/gh4a/widget/LinkSpan.java
Fs00: > I don't see how that scrolling should/could work in our (non-webview) case, so as long as they're present IMHO it's a good idea to let them open the page in the custom tab, scrolled to the right place.

The `rewriteRelativeUrls` function you linked above [doesn't explicitly rewrite anchor links](https://github.com/slapperwan/gh4a/blob/ee5b82386b073b5ccff249d02deba94772bfb193/app/src/main/java/com/gh4a/utils/HtmlUtils.java#L157), since it wouldn't work with the current logic: the anchor should be appended to the README path, not to the repository URL as it does now.

Anyway, since this would be an extra improvement, I think that the guard is enough for now.

Fs00:
> Let me guess: links to external URLs being broken

Nope! That one has been fixed in 07ba31b, hasn't it?
It's a more hidden and subtle issue, you'll see it tomorrow 😉

Fs00:
Pushed the fix for the extra regression (I've added it to the PR description).
You could experience the regression when tapping on "Roadmap" and "Releases" links in [this README](https://github.com/dotnet/runtime/#what-is-net).

PR #1146, opened 2021-11-13
Fixed keeping stale scroll position when switching between fragments.

23rd:
Hi.
When a user switches between accounts in the `News Feed`, a scroll position of their screen stays the same.
It's annoying enough when I have to manually move the scroll back to the original position after switching accounts.
So this small PR fixes that behavior.

It can be reproduced not only with multiple accounts/organizations, but also between `News Feed` and `Public Timeline`.

### DIFF
--- a/app/src/main/java/com/gh4a/fragment/LoadingListFragmentBase.java
+++ b/app/src/main/java/com/gh4a/fragment/LoadingListFragmentBase.java
@@ -132,5 +132,11 @@ public abstract class LoadingListFragmentBase extends LoadingFragmentBase implem
         mFastScroller.setHandlePressedColor(getHighlightColor());
     }
 
+    protected void resetScroll() {
+        if (mLayoutManager != null) {
+            mLayoutManager.scrollToPositionWithOffset(0, 0);
+        }
+    }
+
     protected abstract int getEmptyTextResId();
 }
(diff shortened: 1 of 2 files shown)


maniac103:
Since on user change the fragment is destroyed and a new one is created, I do not yet understand why that new fragment would use state from the old instance. Could you please shed some light on how exactly this happens?

maniac103:
OK, looked it up myself, and I think the patch is wrong. The scroll position is saved via the old fragment saving state, which in turn saves state of all of its views. It's restored when view state is restored, and that makes a lot of sense e.g. on screen rotation, in which case I suspect your patch will cause the list to jump back to the top, which will be very annoying.
To cover the case of account switches one probably should use `RecyclerView.Adapter.setStateRestorationPolicy` to prevent the restoration, but I'm currently wondering what the best way to transmit that information between 'fragment destroy' and 'fragment recreation' time is.

maniac103:
One could probably try to
- overwrite `Fragment.onSaveInstanceState()` and save a hash (or something) of the fragment's arguments to the view state
- overwrite `Fragment.setInitialSavedState()` and only call the super method if the hash in the given state matches the current fragment's arguments

23rd:
Yes, indeed, I did not consider this case, so the first solution was wrong.

I decided to take a different approach. What if we always reset the scroll on non-cache loading of the first page in `PagedDataBaseFragment`? I did some tests and it seems to work better.

Issue #1027, opened 2020-08-31
Jumping to anchors in Markdown files doesn't work


ltGuillaume:
When I click an anchor in a MarkDown file (example: https://github.com/M66B/FairEmail/blob/master/FAQ.md#user-content-known-problems), OctoDroid reopens the entire .md file and remains at the top of the file, instead of jumping to the anchor.

Although I have found https://github.com/slapperwan/gh4a/issues/434, which suggests that the issue has been resolved, OctoDroid still exhibits this behavior.

Using latest OctoDroid from F-Droid on LineageOS 16 (Android 9) with Bromite WebView.

maniac103:
The way the readme mentioned in #434 works fine in OctoDroid, hence that issue is closed.
I assume the readme you mentioned is formatted differently, need to check how exactly.

ltGuillaume:
`<a id="name">` is working in all browsers, but technically `<a name="name">` is the correct way. This is how the .md file I linked to (and many others I found not working in OctoDroid) creates anchors.

PR #1147, opened 2021-11-13
Improve links behavior in Markdown files + rendering tweaks

### LINKED ISSUE #764: GitHub emoji are not rendered in comment preview
Emoji codes (e.g. `:shipit:`) in published issue/PR comments are rendered, but not in the preview when writing a comment.

I'll update this issue if I come across other Markdown elements that are missing from previews.

Is this likely to be something a relative noob to Android development (like me) could tackle? It's Hacktoberfest, after all. :)

### LINKED ISSUE #1090: [Bug] markdown mailto links not working
Example :

```markdown
You can contact me by [email](mailto:foo@bar.com)
```

Expected :
```
mailto:foo@bar.com
```

Actual :
```
https://raw.githubusercontent.com/user-or-org/repo/main/mailto:foo@bar.com
```

Thanks

Fs00:
This PR makes further improvements to the behavior of links especially in Markdown files, which enable a better user experience:
- rewrite relative URLs in Markdown files so that URLs in links become absolute to _github.com_ instead of _raw.github.com_ (fabcdc9). This allows links to repository files to be opened directly in OctoDroid.
- make relative URL rewriting aware of the folder of the current file, so that it works correctly for files that are not in the root of the repo. This makes navigating between MD files linked to each other (such as [dotnet/runtime dev docs](https://github.com/dotnet/runtime/tree/main/docs)) a piece of cake.
- the same URL rewriting logic is now used for both the "Markdown-in-TextView" Readme and for MD files rendered using WebViewerActivity.
  - Since patching showdown.js is not needed anymore, I've also updated the library to the latest stable version (I've used the [minified version](https://github.com/showdownjs/showdown/blob/1.9.1/dist/showdown.min.js)) which also adds the feature requested in #764.
- link click behavior has been made consistent between Markdown-in-TextViews and Markdown-in-WebViews (72f48e4)
- I've fixed a bug in URL rewriting that caused #1090 (274a885) and added explicit handling for `mailto:` links (the introduction of #1118 made them unopenable)

Last but not least, I've made some minor improvements to Markdown-in-WebView styling: better paragraph spacing, fix for invisible separators and some more (4bc9f50). Here's a representative comparison (image was broken since the old URL rewriting in showdown.js didn't work with `` tags in Markdown):

<table>
<tr><td>

</td><td>
</td></tr>
</table>

Fixes #1090 
Closes #764


app/src/main/java/com/gh4a/utils/HtmlUtils.java
```diff
@@ -137,33 +137,31 @@ public void drawBackground(Canvas c, Paint p, int left, int right, int top, int
...
 
-        while (m.find()) {
-            String url = m.group(2);
+        final String baseUrlForImages = "https://raw.github.com/" + repoUser + "/" + repoName + "/" + branch;
+        return rewriteUrlsInAttribute("src", rewrittenHtml, baseUrlForImages);
+    }
+
+    private static String rewriteUrlsInAttribute(String attribute, String html, String baseUrl) {
+        final Matcher matcher = Pattern.compile("(" + attribute + ")=\"(\\S+)\"").matcher(html);
+        final StringBuffer sb = new StringBuffer();
```
maniac103, line 152: Since we have two iterations per rewrite now instead of one, I wonder whether it wouldn't make sense to check whether there's any match at all first, and if there isn't, return `html` right away?

app/src/main/java/com/gh4a/activities/WebViewerActivity.java
```diff
@@ -46,13 +46,16 @@
...
 import com.gh4a.utils.FileUtils;
+import com.gh4a.utils.HtmlUtils;
 import com.gh4a.utils.UiUtils;
 import com.gh4a.widget.FindActionModeCallback;
 import com.gh4a.widget.SwipeRefreshLayout;
 
 import java.io.IOException;
 import java.util.ArrayList;
 
+import static java.lang.String.format;
```
maniac103, line 58: No static imports please.

app/src/main/java/com/gh4a/resolver/LinkClickHandler.java
```diff
@@ -0,0 +1,33 @@
...
+import com.gh4a.utils.IntentUtils;
+
+public class LinkClickHandler {
+    private final BaseActivity mActivity;
+
+    public LinkClickHandler(BaseActivity mActivity) {
+        this.mActivity = mActivity;
+    }
+
+    public void handleClick(Uri clickedUri) {
```
maniac103, line 19: Since this class does not have any internal state, this method can/should very well be `public static void handleClick(BaseActivity activity, Uri clickedUri)` ... and maybe be moved to LinkSpan, just because classes with a single static method IMHO just add clutter without providing too much benefit.

maniac103 (changes requested):

app/src/main/java/com/gh4a/resolver/LinkClickHandler.java
Fs00: Yes, makes sense. Even though calling `LinkSpan.handleClick` would sound a bit confusing when you aren't dealing with actual link spans, at least to me. That's why I extracted the class in the first place.

app/src/main/java/com/gh4a/utils/HtmlUtils.java
Fs00: Then you would have 3 iterations in case you have a match and still one in any case... I don't think there is a concrete performance penalty that justifies a preliminary check, which would still involve a full text scan.

maniac103: Then lazy init `sb` in the first 'found match' run and return `html` instead of `sb` if it's still null.

Fs00: > Then lazy init sb in the first 'found match' run and return html instead of sb if it's still null.

I didn't fully understand what you meant with this, but it seemed to me that the goal was to avoid the copy of the `html` string in the StringBuffer.
I did another thing instead: I've set the initial capacity of the StringBuffer to the length of `html`, to minimize the number of copies/reallocations when appending to the buffer.
I think that any other change that would make the logic more complex is simply not worth it. In the best case you would only save a few memory allocations.

Fs00:
I've made some more adjustments to Markdown-in-WebView rendering:
- tables have now proper font size and collapsed borders
- `inline code snippets` now have appropriate word wrapping so that they won't overflow + cosmetic improvements

Here's two comparisons:

<table>
<tr><td>

</td><td>
</td></tr>
</table>

<table>
<tr><td>

</td><td>
</td></tr>
</table>

Fs00:
I've also found out that the updating Showdown makes Markdown anchor links work in WebView!
May that be considered enough to fix #1027?

maniac103:
> May that be considered enough to fix #1027?

With the example URL in #1027 this doesn't work for me, clicking on the anchor links in that readme does nothing.

Fs00:
> With the example URL in #1027 this doesn't work for me, clicking on the anchor links in that readme does nothing.

Oh, I see why. The anchor links in that Readme have a `user-content-` prefix that works only on GitHub, due to how GH generates the HTML for headings. Here's an example:

```html
<h2 dir="auto">
<a id="user-content-frequently-requested-features" class="anchor" aria-hidden="true" href="#frequently-requested-features">[... omitted ...]</a>
<a name="user-content-frequently-requested-features"></a>Frequently requested features
</h2>
```

The only way to fix this would be not to generate the HTML with Showdown.js, but instead [retrieve it directly from GitHub API](https://docs.github.com/en/rest/reference/repos#custom-media-types-for-repository-contents).

PR #1151, opened 2021-11-17
Update collaborators API parsing

### LINKED ISSUE #927: Error 404 when accessing pull requests of certain repo
When I access any pull request of this repository, even the most recent OctoDroid shows an 404 error, even though these pull requests can be accessed via the GitHub website just fine:
 https://github.com/XboxDev/nxdk

matthiakl:
Fix #927 
See https://docs.github.com/en/rest/reference/repos#check-if-a-user-is-a-repository-collaborator (looks like it has been updated recently):
> Not Found if user is not a collaborator
Status: 404 Not Found


matthiakl:
I just noticed, that  the API returns 404 where I am collaborator, but still 403 if not
 ```
com.gh4a.ApiRequestException: Must have push access to view repository collaborators. (403)
```

maniac103:
Could you ask in [their forum](https://github.community/c/github-ecosystem/github-rest-api/60) for clarification? The 'API returns 404 if user is collaborator' bit doesn't really make sense to me, so it would be good to know whether this is a bug in the API or intentional.

matthiakl:
> Could you ask in their forum for clarification? 

Done. Awaiting review from staff to appear.
[Forum link](https://github.community/t/api-return-codes-for-collaborator-endpoint-does-not-match-reference/213185/1)

matthiakl:
I received no answer from the support, but I found, that the app works as expected, when I use a custom access token instead of OAuth to login.

PR #1154, opened 2021-12-04
Fix a couple of wiki-related bugs

### LINKED ISSUE #1127: [Bug] Direct wiki article links always redirects to "recent wiki updates"

Fs00:
This PR fixes two bugs related to repository wikis:
- links to specific wiki pages incorrectly resolved to the "Recent Wiki Updates" screen (#1127).
Those links now open the wiki page in a browser, since GH doesn't have an API for fetching a wiki page content in order to display it in a WikiActivity.
- app crashed when opening some very large wiki pages from the "Recent Wiki Updates" screen (e.g. https://github.com/ReactiveX/RxJava/wiki/What's-different-in-3.0) due to a TransactionTooLargeException (page content is passed via intent extras).
To fix this, I've made `Feed` implement `Parcelable` and passed it as a whole to the WikiActivity, compressing it as needed. At first I tried to compress/decompress just the `content` intent extra, but I didn't manage to: Android has no public APIs to construct strings from byte arrays (you'd need to use `StringFactory`, which a hidden SDK API).

Fixes #1127

### DIFF
--- a/app/src/main/java/com/gh4a/activities/WikiActivity.java
+++ b/app/src/main/java/com/gh4a/activities/WikiActivity.java
@@ -22,18 +22,21 @@ import androidx.annotation.Nullable;
 
 import com.gh4a.R;
 import com.gh4a.model.Feed;
+import com.gh4a.utils.IntentUtils;
 
 public class WikiActivity extends WebViewerActivity {
     public static Intent makeIntent(Context context, String repoOwner, String repoName, Feed feed) {
-        return new Intent(context, WikiActivity.class)
+        Intent intent = new Intent(context, WikiActivity.class)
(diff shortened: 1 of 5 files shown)


tianalemesle:
> GH doesn't have an API for fetching a wiki page content in order to display it in a WikiActivity

How then do you display a wiki page content from the "recent wiki updates" view ?

maniac103:
There's a feed that has all Wiki page contents, but there's no (official/stable) way to get a single page.

maniac103:
That one is what the Wiki list ('recent wiki updates') is using. The problem is matching the 'page link' part of the URL against the information contained in that feed.

PR #1156, opened 2021-12-21
Do not crash the app on Moshi's JsonDataExceptions

Fs00:
We will never get rid of them (as long as we use the current bindings library). GH can always add new enum values to enums that don't have a null fallback, or there may be enum bindings that are not exhaustive.
Now these errors have a specific error message, so that the user knows what's going on.

Also, I thought it would be a good idea to log all handled exceptions as errors in logcat to make them more prominent.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -29,6 +29,7 @@
     <string name="credentials_otp_prompt">Please enter the two-factor confirmation code you got via SMS.</string>
     <string name="credentials_token">Access token</string>
     <string name="try_again">Try again</string>
+    <string name="report_issue">Report issue</string>
     <string name="ok">OK</string>
     <string name="cancel">Cancel</string>
     <string name="settings">Settings</string>
@@ -68,7 +69,7 @@
     <string name="pub_timeline">Public Timeline</string>
     <string name="load_failure_explanation">Loading the data from GitHub failed. Please verify you have a working network connection and try again.</string>
     <string name="load_failure_explanation_dmca">This repository was taken down for legal reasons.\nPlease see %1$s for details.</string>
-    <string name="load_failure_explanation_parsing">There was an error while parsing GitHub\'s response. See the logcat for details.</string>
+    <string name="load_failure_explanation_parsing">There was an error while parsing GitHub\'s response. You can use the button below to report this issue to the app developers.</string>
     <string name="load_failure_explanation_with_reason">Loading the data from GitHub failed. GitHub told us the following reason:\n%1$s</string>
     <string name="find_next">Find next</string>
     <string name="find_previous">Find previous</string>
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -66,9 +66,10 @@
     <string name="draft">Draft</string>
     <string name="theme">Theme</string>
     <string name="pub_timeline">Public Timeline</string>
-    <string name="load_failure_explanation">Loading the data from Github failed. Please verify you have a working network connection and try again.</string>
+    <string name="load_failure_explanation">Loading the data from GitHub failed. Please verify you have a working network connection and try again.</string>
     <string name="load_failure_explanation_dmca">This repository was taken down for legal reasons.\nPlease see %1$s for details.</string>
-    <string name="load_failure_explanation_with_reason">Loading the data from Github failed. Github told us the following reason:\n%1$s</string>
+    <string name="load_failure_explanation_parsing">There was an error while parsing GitHub\'s response. See the logcat for details.</string>
```
maniac103, line 71: I'm not sure whether showing this message really makes sense with that wording:
- Even though the target audience of OctoDroid is technical, can we reasonably expect all of them to know the term 'logcat'?
- 'logcat' is the log viewer tool, not the log itself ;-)
- Shouldn't we rather tell them to open an issue and attach the log?

For the latter, I think a precondition would be implementing some log collection (spawn logcat process and collect its output into string array).

Fs00: > Shouldn't we rather tell them to open an issue and attach the log?

That's a good idea. I wrote that error message as a less daunting alternative to printing the exception message straight to the user 😅 (which is actually what we care about in those cases).
I think that we don't need to access the log via logcat, given that we have access to the full stack trace. At this point I see two alternatives for the error screen:
- the lazy one: tell the user to report the error and add a button to copy the exception stack trace to the clipboard
- the awesome-UX one: add a button that goes straight to the "New issue" activity for the OctoDroid repo and prepopulates the issue body with a short description and the exception stack trace (maybe slightly overkill?)

What do you think?

maniac103: Given the second option is close to no overhead (since all that needs to be added is a new intent extra in `IssueEditActivity` for title + body), I'm leaning towards the latter. Can repurpose the 'retry' button in that case :-)

Fs00: Done! A couple of notes:
- I haven't pre-filled the issue title for error reports, since this forces the user to try to write a meaningful title for the specific situation
- in the pre-filled issue body I haven't added any device specs, since `JsonDataException`s aren't device or OS-dependent. We can add them later if the "Report issue" features is used in other scenarios
- when the issue report is submitted, we will just display a "Thank you" snackbar (not the created issue). Is this good enough?

maniac103: > when the issue report is submitted, we will just display a "Thank you" snackbar (not the created issue). Is this good enough?

I think so. One could use the snackbar action to go to it ... this would mean we'd need to return e.g. the issue number in the activity result.

Fs00: Done, I've added a 'View' action to the snackbar

maniac103:
Thanks! :+1:

PR #1162, opened 2021-12-27
Add basic support for embedded videos in comments

Fs00:
This PR adds a small QoL life change to allow viewing embedded videos in comments in the app.
Before this PR, you had to open the comment in the GitHub website in order to see an embedded video, because the app displayed only the name of the video file.
Now, a "View video" link will appear near the video name, which opens the video in a custom tab, as you can see in the screenshot below:

I initially thought of making the video file name a clickable link, but it turned out to be too difficult since the `<video>` tag is after the name (which is a `<span>`) in the HTML.

### DIFF
--- a/app/src/main/java/com/gh4a/utils/HtmlUtils.java
+++ b/app/src/main/java/com/gh4a/utils/HtmlUtils.java
@@ -464,6 +464,8 @@ public class HtmlUtils {
                 startHeading(attributes, tag.charAt(1) - '1');
             } else if (tag.equalsIgnoreCase("img")) {
                 startImg(attributes, mImageGetter);
+            } else if (tag.equalsIgnoreCase("video")) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/gh4a/utils/HtmlUtils.java
```diff
@@ -584,6 +586,16 @@ private void appendNewlines(int minNewline) {
             }
         }
 
+        private void appendVideoLink(String videoUrl) {
+            mSpannableStringBuilder.append("\uD83C\uDFA5 "); // movie camera emoji
+            String videoLinkText = "View video";
```
maniac103, line 591: This needs to be a string resource.

Fs00: Oh right, my bad :sweat_smile:

Issue #1132, opened 2021-11-02
Time to raise minSdkVersion?


Fs00:
Currently, this app can be installed on devices down to Android 4.2.
Considering that some notable libraries used by this app (namely OkHttp and RxJava) have dropped support for Android <5, in my opinion it would make sense to aim for a bump to minSdkVersion 21 sooner or later.
It would also bring some neat benefits: no need for multidex and smaller APKs, since bitmap versions of vector drawables (for compatibility with pre-Lollipop devices) won't be bundled in the app.

We could start with a bump to minSdk 19, which should be painless and unlikely to affect more than a handful of users. Outdated [2020 Android distribution data](https://www.xda-developers.com/android-version-distribution-statistics-android-studio/) showed that Android versions prior to 4.4 had the least amount of users (<2%), which should be even less today.

maniac103:
FWIW, our installation data in Google Play as of today:
- Android 11: 2356
- Android 10: 1429
- Android 9: 716
- Unknown (?): 424
- Android 8: 349
- Android 7: 256
- Android 6: 98
- Android 5: 99
- Android 4.4: 48
- Android 4.3: 3
- Android 4.2: 11
- Android 4.1: 1

So yeah, raising to API 19 (4.4) should not be a problem at all, but what will that buy us? AFAICT the real gains only start when bumping to API 21...

Fs00:
Thanks for the info @maniac103.
Do these installation data reflect the number of users that have the app _currently_ installed? That one installation on Android 4.1 seems weird considering that the current minSdk is 17... 🤔 

> So yeah, raising to API 19 (4.4) should not be a problem at all, but what will that buy us? AFAICT the real gains only start when bumping to API 21...

Agree, I've actually thought of the bump to API 19 as a preliminary step for the transition to API 21. We can then bump to API 21 in a subsequent version of the app or just bump from 17 to 21 in a single version. 

A benefit I see in bumping, regardless of the API level, is that we can avoid the potential need for compatibility quirks (speaking of which, I've noticed that WebView on pre-KitKat devices uses a legacy engine with several behavior differences).

What do you think?

maniac103:
> Do these installation data reflect the number of users that have the app currently installed? 

Probably yes, but I believe users that are on older versions would also be counted ... I guess that's the case for that one user.

> Agree, I've actually thought of the bump to API 19 as a preliminary step for the transition to API 21.

I'm not sure why there would be a need for an intermediate step here?
[...]

Issue #857, opened 2018-06-16
Download does not work


dtoxic47:
Trying to download anything frm release page,starts the download but terminates it immediately,trying again the download goes into loop showing two downloads are started,but nothing is beeing dowloaded, my phone is Fujitsu F-02G, Kitkat 4.4.4

dmaulana:
have you give permission to the app to access storage?
are you able to download repo as a zip ?

sick13:
yes all permissions are granted,downloading as a Zip same thing, it goes into endless loop,Downloading "Untitled", no extension or name of the download is given

maniac103:
This seems to be a problem with Kitkat's download manager not being able to talk to Github anymore:

```
W/DownloadManager( 1759): Aborting request for download 10: javax.net.ssl.SSLHandshakeException: javax.net.ssl.SSLProtocolException: SSL handshake aborted: ssl=0xb7ec0550: Failure in SSL library, usually a protocol error
```

I don't think I can do anything about this :-/

Fs00:
@maniac103 I think this can be closed now, given that #1166 has been merged and therefore the app can't run anymore on devices that don't support TLS 1.2.

maniac103:
Right, thanks 👍

PR #1166, opened 2022-01-02
Bump minSdkVersion to 21

Fs00:
As discussed in #1132, this PR bumps the minimum supported Android version to Lollipop (API 21).
The benefits are the following:
- app is ~1.5 MB smaller, thanks to the removal of font assets (which aren't needed since Android 5+ includes all the fonts we need) and raster versions of vector drawables (that were generated at build time for pre-Lollipop compatibility)
- we gain the ability to update some dependencies which do not support anymore Android <5 (namely OkHttp and Retrofit)
- we avoid problems such as #857 caused by Android <5 not always supporting TLS 1.2
- we can leverage more modern CSS/JS features in WebView, some of which will help us to greatly improve maintainability in some areas

Most of the work in this PR has been put into removing custom typeface handling, which resulted in deleting `CustomTypefaceSpan`, ~~`TypefaceCache`, and replacing the uses of `StyleableTextView`'s `ghFont` attribute with `textStyle` and `fontFamily`.~~
~~This makes StyleableTextViews work as one would expect in XML layouts, not ignoring `typeface`/`textStyle`/`fontFamily` attributes.~~


Fs00:
@maniac103 Probably it would be better to split CustomTypefaceSpan/TypefaceCache/ghFont removal in a separate PR?
I understand that the diff is quite big right now.

maniac103:
I guess so, yes, but the main reason I do not want to merge this PR yet is that I think we/I should do a release with all of your fixes merged lately before, so the people with pre-5.0 devices get them.

Fs00:
Makes sense! I've set this PR as draft since it's gonna be paused now.

Just a personal opinion: I would still do a bump to min SDK 19 in any case because personally I don't feel very comfortable continuing to support Android versions that have a radically different WebView implementation, considering also that we did many changes to parts of the app that rely heavily on WebView (and thus I see a potentially higher chance of breaking something on older devices).
WDYT?

maniac103:
>  WDYT?

Done: https://github.com/slapperwan/gh4a/commit/239dde8bcbe44e90e3c5fc4a1c1227fdbf23cd9b

Fs00:
@maniac103 This is now rebased and ready for review.
I've removed from this PR the `TypefaceCache`/`ghFont` attribute removal which was quite an invasive change and led to a high number (and chance) of conflicts.
I've also added a couple of small changes, including the removal of the "pressed" color for the fab, which was conflicting with the ripple effect introduced in API 21.
Despite the high number of "changed" files (many of them have been removed), this PR should be quite easy to review commit-by-commit.

Fs00:
Rebased again to fix a conflict.

Fs00:
@maniac103 How do we want to proceed with this? Given that this PR spans a fair number of files, I think it's best not to delay it further, especially if other big changes like the MD3 redesign are coming soon.
This is also blocking a few other changes I'd like to make in separate PRs.

maniac103:
Since we have done the mentioned release in the mean time, let's ship this.

PR #1173, opened 2022-01-11
More minor fixes and tweaks

Fs00:
Fixed some bugs and applied some tweaks:
- fixed comment deletion not working in diff viewer (the wrong method was being called)
- fixed a bug that caused the repository "Files" tab to misbehave after screen rotation.
When you pressed the back button for the first time after rotation, the path breadcrumbs were updated as if you navigated to the parent folder but the list of contents remained the same. Subsequent back presses updated the list of contents as expected, but at that point the breadcrumbs were out of sync with the directory contents shown.
- revamped issue label selection dialog:
  - fixed padding
  - use standard instead of condensed typeface (it didn't look appropriate to me)
  - shortened _"Manage labels"_ button text to just _"Manage"_, because it was too long to fit in one line on not-enough-large screens :)

- reduced text size from **18sp** to **16sp** in issue/comment editors (feel free to criticize this). In my experience, I've found the former size of 18sp a bit too big when writing long issues or comments.
- minor padding adjustments here and there
- enabled [Java 11 language features](https://developer.android.com/studio/releases/gradle-plugin#java-11) in build.gradle


app/src/main/res/layout/editor_bottom_sheet.xml
```diff
@@ -45,10 +46,12 @@
                 android:id="@+id/et_basic_editor"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
+                android:layout_marginLeft="8dp"
                 android:dropDownAnchor="@id/basic_editor_scroll"
                 android:inputType="textMultiLine|textCapSentences"
-                android:minHeight="48dp"
+                android:minHeight="46dp"
```
maniac103, line 52: Why this change? 48dp is the minimum size of a touchable item in the UI guidelines.

maniac103 (commented):
> minor padding adjustments here and there

Those need some explanation IMHO, especially when deviating from the common/consistent ones.

app/src/main/res/layout/editor_bottom_sheet.xml
Fs00: Didn't know about the guideline.
I've changed it to make the bottom sheet TextView look a bit more centered and aligned with the nearby buttons after reducing text size to 16sp, but the difference is actually very subtle. I'll let you choose :)

app/src/main/res/layout/row_issue_template.xml
```diff
@@ -3,7 +3,7 @@
     android:orientation="vertical"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:paddingHorizontal="@dimen/content_padding"
+    android:paddingHorizontal="24dp"
```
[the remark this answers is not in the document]
Fs00: Because the list elements are not aligned with the dialog title, as you can see below.
I assume that the extra padding is needed because we are using custom views for those elements.

photo1644599852

app/src/main/res/layout/user.xml
```diff
@@ -103,8 +103,7 @@
                     android:layout_width="wrap_content"
                     android:layout_height="wrap_content"
                     android:layout_marginBottom="@dimen/overview_item_spacing"
-                    android:paddingLeft="@dimen/content_padding"
-                    android:paddingRight="@dimen/content_padding"
+                    android:paddingHorizontal="20dp"
```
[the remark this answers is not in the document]
Fs00: I wasn't too sure on this one. It's just that the 16dp padding makes the text look a bit misaligned in some cases, but it depends a lot on how the profile picture looks and on how long the bio text is. 20dp looks better in some cases, but slightly worse in others.
I haven't a strong opinion on this, it's fine for me if we revert it.

app/src/main/res/layout/editor_bottom_sheet.xml
maniac103: I'd prefer to stay with standard sizes, I think.

app/src/main/res/layout/row_issue_create_label.xml
```diff
@@ -21,11 +22,10 @@
         android:layout_height="wrap_content"
         android:foreground="?attr/selectableItemBackground"
         android:paddingBottom="5dp"
-        android:paddingLeft="8dp"
+        android:paddingLeft="16dp"
         android:paddingRight="8dp"
         android:paddingTop="5dp"
         android:textAppearance="?android:attr/textAppearanceMedium"
-        app:ghFont="condensed"
```
maniac103, line 28: With that change the label selection UI is now inconsistent to the label management UI ... I'd prefer to keep this.

app/src/main/res/layout/user.xml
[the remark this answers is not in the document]
maniac103: Then please keep the common padding :+1:

app/src/main/res/layout/editor_bottom_sheet.xml
Fs00: For "stay with standard sizes" you mean only for the `minHeight` or for both `textSize` and `minHeight`?

app/src/main/res/layout/row_issue_create_label.xml
Fs00: I don't really agree on this, but I don't mind reverting this change if you have a strong opinion.
To keep consistency with the other screen, I'd also consider changing the label management UI not to use condensed font, which together with list separators gives it a quite dated look IMO.

app/src/main/res/layout/editor_bottom_sheet.xml
maniac103: I was referring to `minHeight` only.

app/src/main/res/layout/row_issue_create_label.xml
maniac103: As long as both are consistent, I don't have an overly strong opinion either way.

Fs00: Reverted this change for now to keep UIs consistent and unblock the other changes in this PR

PR #1175, opened 2022-01-21
add data loss functionality to issue fragment base

Naitim:
Hello developers of octodroid

I am using your app octodroid. I think the app is great but I have one minor patch that could improve the user experience.

Here is a picture to help illustrate what activity that are changed in my patch:

https://ibb.co/6vgx1S3

When the user is typing a message to send. If the screen focus goes to another app or activity or if the user accidentally closes or press back the app, the user will lose any data they had put into this page

This feature will automatically store the data when the user leaves the activity without submitting and restore said data when they restart the activity(this can be seen here where I pressed back and went back into the activity https://ibb.co/gPL8BNB). Therefore, the user does not have to fill in the data again thus improving the user experience.

I have tested out the feature to ensure it works.

If you have any questions or if you would like me to change anything, please do not hesitate to let me know!

Thank you for your time,
Tim

### DIFF
--- a/app/src/main/java/com/gh4a/fragment/IssueFragmentBase.java
+++ b/app/src/main/java/com/gh4a/fragment/IssueFragmentBase.java
@@ -17,6 +17,7 @@ package com.gh4a.fragment;
 
 import android.app.Activity;
 import android.content.Intent;
+import android.content.SharedPreferences;
 import android.graphics.Typeface;
 import android.os.Bundle;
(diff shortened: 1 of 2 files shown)


RSoulwin:
@maniac103 @Fs00
What are views of devs about this PR being added ?
has it been reviewed? 
..or forgotten as it is there for quite a while in repo but i think it has a useful feature from user's prospective.

PR #1182, opened 2022-02-19
Improve clarity for `referenced` and `cross-referenced` timeline events

Fs00:
This PR makes the following improvements to the way `referenced` and `cross-referenced` timeline events are displayed:
- `referenced`: display also repository owner and name together with the SHA of the commit that referenced the issue/PR, if it belongs to another repository.
I've also reworded the message for the event since I've discovered that references from commit comments (and not only from commit messages) appear as `referenced` events too (see [this](https://github.com/PowerShell/PowerShell/pull/16840#ref-commit-f8168cb) for an example), and therefore the message _"added commit [] that referenced this issue"_ could have been misleading.
- `cross-referenced`: display also repository owner and name together with the number of the issue that referenced the issue/PR, if it belongs to another repository (like GitHub does, e.g. `dotnet/runtime#12456`).

I've also extracted a few methods to improve code readability.

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
+++ b/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
@@ -262,24 +262,31 @@ class EventViewHolder
 
         int pos = text.toString().indexOf("[commit]");
         if (event.commitId() != null && pos >= 0) {
-            text.replace(pos, pos + 8, event.commitId().substring(0, 7));
-            text.setSpan(new IntentSpan(mContext, context -> {
-                // The commit might be in a different repo. The API doesn't provide
-                // that information directly, so get it indirectly by parsing the URL
-                String repoOwner = mRepoOwner;
-                String repoName = mRepoName;
-                String url = event.commitUrl();
-                if (url != null) {
-                    Matcher matcher = COMMIT_URL_REPO_NAME_AND_OWNER_PATTERN.matcher(url);
-                    if (matcher.find()) {
-                        repoOwner = matcher.group(1);
-                        repoName = matcher.group(2);
-                    }
+            // The commit might be in a different repo. The API doesn't provide
+            // that information directly, so get it indirectly by parsing the URL
+            String repoOwner = mRepoOwner;
+            String repoName = mRepoName;
+            boolean isInDifferentRepo = false;
+            String url = event.commitUrl();
+            if (url != null) {
+                Matcher matcher = COMMIT_URL_REPO_NAME_AND_OWNER_PATTERN.matcher(url);
+                if (matcher.find()) {
+                    repoOwner = matcher.group(1);
+                    repoName = matcher.group(2);
+                    isInDifferentRepo = !mRepoOwner.equals(repoOwner) || !mRepoName.equals(repoName);
                 }
+            }
+            String shortCommitSha = event.commitId().substring(0, 7);
+            String commitText = isInDifferentRepo
+                    ? repoOwner + "/" + repoName + "#" + shortCommitSha
+                    : shortCommitSha;
+            text.replace(pos, pos + 8, commitText);
 
-                return CommitActivity.makeIntent(context, repoOwner, repoName, event.commitId());
-            }), pos, pos + 7, 0);
-            text.setSpan(new TypefaceSpan("monospace"), pos, pos + 7, 0);
+            String finalRepoOwner = repoOwner;
+            String finalRepoName = repoName;
+            text.setSpan(new IntentSpan(mContext, context ->
+                    CommitActivity.makeIntent(context, finalRepoOwner, finalRepoName, event.commitId())), pos, pos + commitText.length(), 0);
+            text.setSpan(new TypefaceSpan("monospace"), pos, pos + commitText.length(), 0);
         }
 
         pos = text.toString().indexOf("[label]");
--- a/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
+++ b/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
@@ -306,13 +306,19 @@ class EventViewHolder
         pos = text.toString().indexOf("[source]");
         if (pos >= 0) {
             final Issue source = event.source().issue();
-            String sourceLabel = "#" + source.number();
+            var sourceRepoOwnerAndName = ApiHelpers.extractRepoOwnerAndNameFromIssue(source);
+            String sourceRepoOwner = sourceRepoOwnerAndName.first;
+            String sourceRepoName = sourceRepoOwnerAndName.second;
+            boolean isSourceInDifferentRepo = !mRepoOwner.equals(sourceRepoOwner) || !mRepoName.equals(sourceRepoName);
+            String sourceLabel = isSourceInDifferentRepo
+                    ? sourceRepoOwner + "/" + sourceRepoName + "#" + source.number()
+                    : "#" + source.number();
             text.replace(pos, pos + 8, sourceLabel);
-            text.setSpan(new IntentSpan(mContext, context -> {
-                return source.pullRequest() != null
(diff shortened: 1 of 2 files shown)


maniac103:
In general this LGTM, but...

> I've also extracted a few methods to improve code readability.

... I have to admit I'm not a fan of such one-off methods, because it makes one jump back-and-forth through the code when following the logic.
Since each of those pieces of code is in its own block due to the `indexOf` check there's no improvement in variable scoping, and there's no dependency between them so I honestly don't see the improvement here.

Fs00:
I think there are a few compelling reasons to extract those sections in separate methods:

- the `formatEvent()` method is quite daunting (220 lines of code before the refactoring) and it's hard to get an overview of what's doing by just peeking at the code.
- when I first looked at the code, I had to spend some time parsing a few of those sections to understand that they were substituting a placeholder in the text. By extracting them and giving them a name, it's easier to understand what they are doing without having to get into the details of the logic.
- *"there's no dependency between them"*: true, and it sounds like a very good reason to extract them to me, because in this way you enforce the decoupling but most importantly you make it very explicit (it's not very visible when you read the code the first time).

About the jumping back-and-forth, and I don't think that's the case here. The details of those sections aren't really important to know if you're just following the logic, and furthermore it seems to me that the sections are changed independently of each other in most cases (so it's unlikely you'll need to jump around), and having each of them extracted helps (me at least) focusing on one piece at a time.
Also, you can see from the diff of commit 6d188a8 that the code wasn't moved around much, the main changes are the method declarations being added in between the code.

Just to be clear, I don't want to impose my way of structuring code here :sweat_smile:, my goal was just to avoid to others the struggle I've had when first approaching the code.
It's also fine for me if you prefer to revert the refactoring.

maniac103:
How about adding comments as a (meaningful) compromise?

Fs00:
Definitely better than nothing for someone reading the code for the first time, but you won't get the aforementioned benefits about explicit scoping, "hiding" unnecessary details when just following the logic, and making the method size more manageable.
Also, I personally prefer when the code itself tells what it's doing, rather than the human using comments (unless when needed to explain motivations or reasonings behind the code).

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
```diff
@@ -258,69 +258,96 @@ private CharSequence formatEvent(final IssueEvent event, final User user, int ty
...
+        replaceLabelPlaceholder(event, text);
+        replaceBotPlaceholder(text);
+        replaceSourcePlaceholder(event, text);
+        replaceTimePlaceholder(event, text);
+        return text;
+    }
 
+    private void replaceCommitPlaceholder(IssueEvent event, SpannableStringBuilder text) {
         int pos = text.toString().indexOf("[commit]");
         if (event.commitId() != null && pos >= 0) {
```
maniac103, line 273: Please (at least ;-) ) invert the logic here and return early if preconditions aren't met, so indention depth is reduced now we don't need the block for variable scoping anymore:

```
if (pos < 0 || event.commitId() == null) {
    return;
}
```

(same below)

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
```diff
@@ -258,69 +258,96 @@ private CharSequence formatEvent(final IssueEvent event, final User user, int ty
...
+                            : IssueActivity.makeIntent(context, sourceRepoOwner, sourceRepoName, source.number())
+            ), pos, pos + sourceLabel.length(), 0);
         }
+    }
 
-        CharSequence time = event.createdAt() != null
-                ? StringUtils.formatRelativeTime(mContext, event.createdAt(), true) : "";
-
-        pos = text.toString().indexOf("[time]");
+    private void replaceTimePlaceholder(IssueEvent event, SpannableStringBuilder text) {
```
maniac103, line 340: Pass the Date instance instead of the full event?

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
```diff
@@ -258,69 +258,96 @@ private CharSequence formatEvent(final IssueEvent event, final User user, int ty
...
+    private void replaceBotPlaceholder(SpannableStringBuilder text) {
+        int pos = text.toString().indexOf("[bot]");
         if (pos >= 0) {
             text.delete(pos, pos + 5);
             StringUtils.addUserTypeSpan(mContext, text, pos, mContext.getString(R.string.user_type_bot));
         }
+    }
 
-        pos = text.toString().indexOf("[source]");
+    private void replaceSourcePlaceholder(IssueEvent event, SpannableStringBuilder text) {
```
maniac103, line 320: Pass source instead of event?

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
```diff
@@ -258,69 +258,96 @@ private CharSequence formatEvent(final IssueEvent event, final User user, int ty
...
+            String finalRepoOwner = commitRepoOwner;
+            String finalRepoName = commitRepoName;
+            text.setSpan(new IntentSpan(mContext, context ->
+                    CommitActivity.makeIntent(context, finalRepoOwner, finalRepoName, event.commitId())), pos, pos + commitText.length(), 0);
+            text.setSpan(new TypefaceSpan("monospace"), pos, pos + commitText.length(), 0);
         }
+    }
 
-        pos = text.toString().indexOf("[label]");
+    private void replaceLabelPlaceholder(IssueEvent event, SpannableStringBuilder text) {
```
maniac103, line 302: Pass label instead of event?

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
Fs00: Here's another benefit of extracting methods, the ability to return early :wink: 
Will do it in a moment

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
```diff
@@ -258,69 +259,94 @@ private CharSequence formatEvent(final IssueEvent event, final User user, int ty
...
+        ), pos, pos + sourceLabel.length(), 0);
+    }
 
-        CharSequence time = event.createdAt() != null
-                ? StringUtils.formatRelativeTime(mContext, event.createdAt(), true) : "";
-
-        pos = text.toString().indexOf("[time]");
+    private void replaceTimePlaceholder(SpannableStringBuilder text, Date time) {
+        int pos = text.toString().indexOf("[time]");
         if (pos >= 0) {
```
maniac103, line 343: Another place for early return.

app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
```diff
@@ -258,69 +259,94 @@ private CharSequence formatEvent(final IssueEvent event, final User user, int ty
...
+        // that information directly, so get it indirectly by parsing the URL
+        String commitRepoOwner = mRepoOwner;
+        String commitRepoName = mRepoName;
+        boolean isCommitInDifferentRepo = false;
+        if (commitUrl != null) {
+            Matcher matcher = COMMIT_URL_REPO_NAME_AND_OWNER_PATTERN.matcher(commitUrl);
+            if (matcher.find()) {
+                commitRepoOwner = matcher.group(1);
+                commitRepoName = matcher.group(2);
+                isCommitInDifferentRepo = !mRepoOwner.equals(commitRepoOwner) || !mRepoName.equals(commitRepoName);
```
maniac103, line 287: Can do this outside of the if below (or even get rid of the variable completely).

Fs00: Done, I've kept the variable since it makes the condition more readable IMO.

PR #1185, opened 2022-03-04
Optimize image loading by avoiding redundant file read/writes

Fs00:
While I was taking a look at how the app fetches images in comments and Readmes, I was surprised to discover that before rendering, images were being written to a temporary file just to be read from it immediately after.
Among a couple of small refactorings, the PR changes the logic to avoid this unnecessary and inefficient step, which also spares us quite a bit of boilerplate error handling code.

### DIFF
--- a/app/src/main/java/com/gh4a/utils/HttpImageGetter.java
+++ b/app/src/main/java/com/gh4a/utils/HttpImageGetter.java
@@ -328,8 +328,8 @@ public class HttpImageGetter {
 
     private final Context mContext;
 
-    private final int mWidth;
-    private final int mHeight;
+    private final int mMaxWidth;
+    private final int mMaxHeight;
 
     private boolean mDestroyed;
 
@@ -339,10 +339,9 @@ public class HttpImageGetter {
 
         WindowManager wm = (WindowManager) context.getSystemService(Context.WINDOW_SERVICE);
         final Point size = new Point();
-
         wm.getDefaultDisplay().getSize(size);
-        mWidth = size.x;
-        mHeight = size.y;
+        mMaxWidth = size.x;
+        mMaxHeight = size.y;
 
         mLoadingDrawable = ContextCompat.getDrawable(context, R.drawable.image_loading);
         mLoadingDrawable.setBounds(0, 0,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/gh4a/utils/HttpImageGetter.java
```diff
@@ -440,14 +436,13 @@ private Drawable loadImageForUrl(String source) {
...
-            InputStream is = null;
             Request request = new Request.Builder()
                     .url(url)
                     .build();
             try (Response response = mClient.newCall(request).execute()) {
-                is = response.body().byteStream();
-                if (is != null) {
+                if (response.body() != null) {
+                    byte[] responseBody = response.body().bytes();
+                    InputStream is = new ByteArrayInputStream(responseBody);
```
Fs00, line 445: An interesting sidenote: using a `ByteArrayInputStream` here instead of OkHttp `byteStream` makes `URLConnection.guessContentTypeFromStream` (used at line 452) actually work, since OkHttp streams do not support marking which is required by the latter function to work.

maniac103: That fact probably deserves a comment.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
```diff
@@ -457,36 +452,23 @@ private Drawable loadImageForUrl(String source) {
...
-                            output = File.createTempFile("image", ".tmp", mCacheDir);
-                            if (FileUtils.save(output, is)) {
-                                if (isGif) {
-                                    GifDrawable d = new GifDrawable(output);
-                                    d.setBounds(0, 0, d.getIntrinsicWidth(), d.getIntrinsicHeight());
-                                    return d;
-                                } else {
-                                    bitmap = getBitmap(output, mWidth, mHeight);
-                                }
+                            if (isGif) {
```
maniac103, line 459: Can probably save one indention level now:
```
if (isGif && canLoadGif()) {
    ...
} else if (!isGif) {
    ...
}

maniac103:
> I was surprised to discover that before rendering, images were being written to a temporary file just to be read from it immediately after.

I think that's fully understandable. I wonder why I never questioned it in all those years ... after all, this code wasn't written by us, but [imported](https://github.com/slapperwan/gh4a/commit/6af7792aa1057136121b87509c81a7a8be181d57) straight from the old GH Android app.

Fs00:
> after all, this code wasn't written by us, but imported straight from the old GH Android app.

I discovered that too, while digging into the history of the file to see if there was a particular reason for that implementation (apparently not).
You often find interesting things in the history of a 10-year-old project :)

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
Fs00: Done. The condition is now much more easily understandable IMO

PR #1188, opened 2022-03-16
A couple of small MIME type detection fixes

### LINKED ISSUE #1177: .mjs file issues
Files that use the `.mjs` extension open in an external window instead of opening in the code editor.

Fs00:
This PR includes the following fixes:
- .mjs files (ES Modules) won't be recognized as binary files anymore on Android 11+ (#1177) and will therefore be opened in the in-app code viewer
- when querying the MIME type overrides map, we were assuming that the extension was always lowercased, and therefore a `.bat` file would have been opened in-app whereas a `.BAT` file would have been opened in the browser.
We now always use the lowercased extension to avoid this issue.

Fixes #1177

### DIFF
--- a/app/src/main/java/com/gh4a/utils/FileUtils.java
+++ b/app/src/main/java/com/gh4a/utils/FileUtils.java
@@ -120,10 +120,11 @@ public class FileUtils {
         if (StringUtils.isBlank(extension)) {
             return null;
         }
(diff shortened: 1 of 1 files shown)


maniac103:
Thanks!

seirdotexe:
Thanks for this one! @maniac103 is it possible a new version can be released on the Play Store?

PR #1205, opened 2022-05-21
Improve text sizes in Markdown rendering

Fs00:
I've made a couple of adjustments regarding text sizes to Markdown rendering in comments/Readme:
- <sub>subscript</sub>/<sup>superscript</sup> text is now smaller, to make it look as it does in GitHub web UI
- sizes of H1 to H6 has been changed to make the different types of heading more clearly distinguishable. Previously I had a hard time finding the difference between an H1 and an H2, or between an H2 and an H3. You can see below the difference:
<table>
<tr><td>

</td><td>
</td></tr>
</table>

### DIFF
--- a/app/src/main/java/com/gh4a/utils/HtmlUtils.java
+++ b/app/src/main/java/com/gh4a/utils/HtmlUtils.java
@@ -232,7 +232,7 @@ public class HtmlUtils {
 
     private static class HtmlToSpannedConverter implements ContentHandler {
         private static final float[] HEADING_SIZES = {
-            1.5f, 1.4f, 1.3f, 1.2f, 1.1f, 1f,
+            1.75f, 1.5f, 1.25f, 1.1f, 1f, 0.9f
         };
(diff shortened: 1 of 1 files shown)


Fs00:
@maniac103 Can this be merged too? Any doubts?

maniac103:
The only thing I have doubts about is that I'm not sure about H6 size being smaller than normal text.

Fs00:
I thought it was fine because I've seen that the GitHub UI makes H6 even smaller than we do (on the GH website H6 are 85% of the normal text size compared to 90% in the app).

An alternative I see is to make H6 as big as normal text and H5, but change their color to a dark shade of grey (both the GitHub UI and [our Markdown CSS](https://github.com/slapperwan/gh4a/blob/master/app/src/main/assets/markdown-light.css#L60) use this approach).

maniac103:
If they're doing the same thing it's fine I guess.

PR #1223, opened 2022-08-16
Display issue closed reason in timeline events

Fs00:
This PR keeps OctoDroid up with the [changes GitHub has recently made to the "closed" issue status](https://github.com/github/roadmap/issues/289):
- the "closed as completed" and "closed as not planned" information has been added to the `closed` timeline event
  
- ~~the IssueActivity/Fragment color has been changed to purple if the issue is _closed as completed_.~~
~~For _closed as not planned_ issues I've kept the classic red color - even if GitHub UI uses dark grey - for consistency with the closed red icon that is already used in the app, and because the grey color used by GH is very similar to the grey used for PR draft status (and draft issues may be a thing in the future, according to the issue linked above)~~
- ~~the Closed tab in the Issues list has had its color changed from red to purple (since issues are by default _closed as completed_)~~
  - ~~following the same logic, the color of the _My issues_ list has been changed to purple when showing closed issues~~
- small extra: I've added the reason why the issue has been locked in the `locked` timeline event when available

I haven't added the ability to close an issue as not planned since it seems it can't be done via the REST API: we need to use the [closeIssue GraphQL mutation](https://docs.github.com/en/graphql/reference/mutations#closeissue). Also, when we introduce that feature, the UI will need to be revised since it's possible to change the closed reason of an issue by "re-closing" it.

**A few more technical details that will be useful while reviewing this:**
- in timeline events, the `stateReason` field is null if the issue was closed before the introduction of closed reasons. The GitHub UI displays the event with _closed as completed_ when `stateReason` is `completed` or `null`
- the Issue `stateReason` field is never null for closed issues: it's set to `completed` for both _closed as completed_ issues and issues that were closed before the introduction of closed reasons

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
+++ b/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
@@ -198,7 +198,12 @@ class EventViewHolder
                 textResId = R.string.issue_event_unlabeled;
                 break;
             case Locked:
-                textResId = R.string.issue_event_locked;
+                if (event.lockReason() == null) {
+                    textResId = R.string.issue_event_locked;
+                } else {
+                    textBase = mContext.getString(R.string.issue_event_locked_with_reason,
+                        getUserLoginWithBotSuffix(user), event.lockReason());
+                }
                 break;
             case Unlocked:
                 textResId = R.string.issue_event_unlocked;
--- a/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
+++ b/app/src/main/java/com/gh4a/adapter/timeline/EventViewHolder.java
@@ -26,6 +26,7 @@ import com.gh4a.widget.TimestampToastSpan;
 import com.meisolsson.githubsdk.model.Issue;
 import com.meisolsson.githubsdk.model.IssueEvent;
 import com.meisolsson.githubsdk.model.IssueEventType;
+import com.meisolsson.githubsdk.model.IssueStateReason;
 import com.meisolsson.githubsdk.model.Label;
 import com.meisolsson.githubsdk.model.Rename;
 import com.meisolsson.githubsdk.model.User;
@@ -33,6 +34,7 @@ import com.meisolsson.githubsdk.model.User;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/gh4a/activities/IssueListActivity.java
```diff
@@ -114,15 +114,20 @@ public static Intent makeIntent(Context context, String repoOwner, String repoNa
...
 
-    private static final int[] TITLES = new int[] {
+    private static final int[] ISSUE_TITLES = new int[] {
         R.string.open, R.string.closed
     };
 
+    private static final int[][] ISSUE_HEADER_COLOR_ATTRS = new int[][] {
+        { R.attr.colorIssueOpen, R.attr.colorIssueOpenDark },
+        { R.attr.colorIssueClosedCompleted, R.attr.colorIssueClosedCompletedDark }
+    };
```
maniac103, line 124: Since the issue activity distinguishes between 'open', 'closed' and 'completed' now, I think the issue list activity should do the same.

maniac103 (changes requested):

app/src/main/java/com/gh4a/activities/IssueListActivity.java
Fs00: I don't agree on this... "completed" is the reason why an issue has been closed, it isn't a third state.
Furthermore, if we follow this logic, the three tabs should be called "Open",  "Not planned" and "Completed". But even in that way it would look overkill IMHO, considering how the GitHub UI displays the new close reasons.

maniac103: 'Completed' basically is the equivalent of the 'merged' PR state: 'closed as done' ... treating it that way in the UI doesn't seem like overkill to me, but rather more consistent?

Fs00: > 'Completed' basically is the equivalent of the 'merged' PR state: 'closed as done'

Mmhh... I believe it's a bit of a stretch to see "closed as completed" in the same way as the Merged state.
The GitHub UI emphasizes it differently (there's no "Completed" purple pill for issues like the Merged pill for PRs) and there's also the fact that all old issues are displayed as "closed as completed" while they might have been closed for other reasons... this is why I don't like having a "hard" separation in the UI.
My idea in this PR was to replicate the GH web UI behavior in the app, with the only difference that _not planned_ is red instead of grey.

> show all closed issues, with red header

I guess it could be fine, but what about using purple for all issues instead? Red is not used anymore for issues in GitHub UI and _completed_ has become the "default" closed reason.

Fs00:
@maniac103 I've removed the commits that change the header colors, can this be merged as-is for now?

PR #1224, opened 2022-08-17
Add a setting to disable custom tabs

the-blank-x:
https://developer.chrome.com/docs/android/custom-tabs/best-practices/#allow-the-user-to-opt-out-of-custom-tabs

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -592,4 +592,8 @@
     <string name="theme_light">Day</string>
     <string name="theme_dark">Night</string>
     <string name="theme_system">Use system setting</string>
+
+    <string name="use_custom_tabs">Use built-in browser</string>
+    <string name="open_links_inside_app">Open links inside the app</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -592,4 +592,6 @@
     <string name="theme_light">Day</string>
     <string name="theme_dark">Night</string>
     <string name="theme_system">Use system setting</string>
+
+    <string name="use_custom_tabs">Use custom tabs</string>
```
maniac103, line 596: I don't like this string, because I'm not sure the majority of the users knows the term 'custom tabs'. Other apps I use (namely Feedly and Discord) offer a selection (`ListPreference`) named along the lines of `Used browser type` with the options `Built-in` and `External`.
One could keep the `SwitchPreference`, give it the title `Use built-in browser` and use the summaries `Open links inside the app` (on state) and `Open links in external browser` (off state).

maniac103 (changes requested):

Issue #1228, opened 2022-08-29
Does not jump/scroll to first unread post


ltGuillaume:
This used to (mostly) work, now it's completely failing: when opening a known issue, OctoDroid stays at the top (first post), instead of jumping/scrolling to the first unread post.

OctoDroid 4.6.8 F-Droid
Android 11

Fs00:
> when opening a known issue

You mean opening it from the Notifications screen?

ltGuillaume:
Yes!

Issue #861, opened 2018-08-13
Remove swipe-to-remove on the bookmarks screen


smichel17:
Occasionally it is useful, but when I actually want to remove a bookmark, it's not very much harder to just go to the repo, like if I wanted to unstar something.

On the other hand, especially now that `bookmarks and stars` are tabbed, I accidentally swipe away a bookmark much more often

badrievafatima5:
Hello. In a chat you can learn a lot of useful information.

PR #1230, opened 2022-08-31
Miscellanous quality-of-life changes

Fs00:
This PR is centered around small improvements to the user experience and minor usability fixes:

- placeholder icons for images that aren't loaded in comments/READMEs have been changed: now they differ when an image can't be loaded or when a GIF is not loaded because of the loading preferences (previously the same placeholder was used)
<table>
<tr><td>

</td><td>
</td></tr>
</table>

- in code search results, the full path to the matching file is shown instead of just the file name
- the scaling logic for images in comments/READMEs has been adjusted, so that images that are taller than the screen height are not uselessly resized when they aren't wider than the screen (the diagram in [TeamNewPipe/NewPipe#7673](https://github.com/TeamNewPipe/NewPipe/issues/7673) is a good example).
While I was at it, I've noticed that the scaling of SVGs whose size is not in pixels was incorrect and I've fixed it (1d60a97)
- bookmarks cannot be removed by swiping them to the left anymore, to avoid the usability issue described in #861.
They can still be removed by swiping to the right.
- last but not least, I've improved the logic to jump to the most recent comment in an issue/PR when clicking on a notification (addresses #1228):
  - we now support scrolling down to the most recent _event_ too (it doesn't need to be a comment or a review anymore)
  - we now look at the `updated_at` field of the notification and not at `last_read_at` to determine the timestamp of the latest event. The latter field is updated once the notification is read, and therefore tapping a notification just after reading it would make the user jump to a different position (most of the time it wouldn't jump anywhere, because there would be no events matching the `last_read_at` timestamp)

### DIFF
--- a/app/src/main/java/com/gh4a/adapter/SearchAdapter.java
+++ b/app/src/main/java/com/gh4a/adapter/SearchAdapter.java
@@ -93,7 +93,7 @@ public class SearchAdapter extends RootAdapter<Object, RecyclerView.ViewHolder> 
         @Override
         public void onBindViewHolder(ViewHolder holder, SearchCode result) {
             Repository repo = result.repository();
-            holder.tvTitle.setText(result.name());
+            holder.tvTitle.setText(result.path());
             holder.tvRepo.setText(ApiHelpers.formatRepoName(mContext, repo));
 
             List<TextMatch> matches = result.textMatches();
@@ -127,6 +127,7 @@ public class SearchAdapter extends RootAdapter<Object, RecyclerView.ViewHolder> 
                     tvMatch.setText(builder);
                     tvMatch.setTag(result);
                     tvMatch.setTag(R.id.search_match_index, i);
+                    tvMatch.setTypeface(Typeface.MONOSPACE);
                     row.setVisibility(View.VISIBLE);
                 }
                 for (int i = matches.size(); i < holder.matchesContainer.getChildCount(); i++) {
--- a/app/src/main/java/com/gh4a/utils/HttpImageGetter.java
+++ b/app/src/main/java/com/gh4a/utils/HttpImageGetter.java
@@ -323,6 +323,7 @@ public class HttpImageGetter {
 
     private final Handler mHandler = new Handler();
     private final Map<Object, ObjectInfo> mObjectInfos = new HashMap<>();
+    private final Drawable mGifPlaceholderDrawable;
     private final Drawable mLoadingDrawable;
     private final Drawable mErrorDrawable;
     private final OkHttpClient mClient;
@@ -344,11 +345,15 @@ public class HttpImageGetter {
         mMaxWidth = size.x;
         mMaxHeight = size.y;
 
+        mGifPlaceholderDrawable = ContextCompat.getDrawable(context, R.drawable.image_gif_placeholder);
+        mGifPlaceholderDrawable.setBounds(0, 0,
+                mGifPlaceholderDrawable.getIntrinsicWidth(), mGifPlaceholderDrawable.getIntrinsicHeight());
+
         mLoadingDrawable = ContextCompat.getDrawable(context, R.drawable.image_loading);
         mLoadingDrawable.setBounds(0, 0,
                 mLoadingDrawable.getIntrinsicWidth(), mLoadingDrawable.getIntrinsicHeight());
 
-        mErrorDrawable = ContextCompat.getDrawable(context, R.drawable.content_picture);
+        mErrorDrawable = ContextCompat.getDrawable(context, R.drawable.image_error);
         mErrorDrawable.setBounds(0, 0,
                 mErrorDrawable.getIntrinsicWidth(), mErrorDrawable.getIntrinsicHeight());
     }
@@ -457,11 +462,15 @@ public class HttpImageGetter {
                         bitmap = renderSvgToBitmap(mContext.getResources(), is);
                     } else {
                         boolean isGif = mime != null && mime.startsWith("image/gif");
-                        if (isGif && canLoadGif()) {
-                            GifDrawable d = new GifDrawable(responseBody);
-                            d.setBounds(0, 0, d.getIntrinsicWidth(), d.getIntrinsicHeight());
-                            return d;
-                        } else if (!isGif) {
+                        if (isGif) {
+                            if (canLoadGif()) {
+                                GifDrawable d = new GifDrawable(responseBody);
+                                d.setBounds(0, 0, d.getIntrinsicWidth(), d.getIntrinsicHeight());
+                                return d;
+                            } else {
+                                return mGifPlaceholderDrawable;
+                            }
+                        } else {
                             bitmap = getBitmap(responseBody);
                         }
                     }
--- a/app/src/main/java/com/gh4a/utils/HttpImageGetter.java
+++ b/app/src/main/java/com/gh4a/utils/HttpImageGetter.java
@@ -517,7 +517,7 @@ public class HttpImageGetter {
         BitmapFactory.decodeByteArray(image, 0, image.length, options);
 
         int scale = 1;
-        while (options.outWidth >= mMaxWidth || options.outHeight >= mMaxHeight) {
+        while (options.outWidth >= mMaxWidth) {
             options.outWidth /= 2;
             options.outHeight /= 2;
             scale *= 2;
@@ -561,7 +561,7 @@ public class HttpImageGetter {
                     density = null;
                 }
 
-                while (docWidth >= mMaxWidth || docHeight >= mMaxHeight) {
+                while (docWidth >= mMaxWidth) {
                     docWidth /= 2;
                     docHeight /= 2;
                     if (density != null) {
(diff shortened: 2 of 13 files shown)


app/src/main/java/com/gh4a/adapter/SearchAdapter.java
```diff
@@ -127,6 +127,7 @@ public void onBindViewHolder(ViewHolder holder, SearchCode result) {
                     tvMatch.setText(builder);
                     tvMatch.setTag(result);
                     tvMatch.setTag(R.id.search_match_index, i);
+                    tvMatch.setTypeface(Typeface.MONOSPACE);
```
maniac103, line 130: Can this be done in XML?

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
```diff
@@ -534,7 +534,6 @@ private Bitmap renderSvgToBitmap(Resources res, InputStream is) {
         try {
             SVG svg = SVG.getFromInputStream(is);
             if (svg != null) {
-                svg.setRenderDPI(DisplayMetrics.DENSITY_DEFAULT);
```
maniac103, line 528: Please explain the exact problem. I do not remember details, but I *do* remember this line was part of a lot of fiddling around to get SVGs render in correct size in all cases.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
```diff
@@ -342,8 +340,7 @@ public HttpImageGetter(Context context) {
         WindowManager wm = (WindowManager) context.getSystemService(Context.WINDOW_SERVICE);
         final Point size = new Point();
         wm.getDefaultDisplay().getSize(size);
-        mMaxWidth = size.x;
-        mMaxHeight = size.y;
+        mMaxWidth = size.x - UiUtils.dpToPixels(context, 48);  // take horizontal padding into account
```
maniac103, line 343: Do not hardcode sizes, for two reasons:
- the image getter doesn't know about the view it's used in
- different views might use different paddings

If anything, pass in the actual view's usable size, but since the point of scaling down beyond the screen size is just to save memory, I'm not sure what the point of that is?

maniac103 (changes requested):

app/src/main/java/com/gh4a/adapter/SearchAdapter.java
Fs00: Unfortunately no, because at the moment StyleableTextView doesn't handle the `typeface` attribute correctly. (#1166 will fix that)

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
```diff
@@ -538,32 +535,25 @@ private Bitmap renderSvgToBitmap(Resources res, InputStream is) {
...
-                        float heightForAspect = (float) mMaxWidth / aspectRatio;
-                        float widthForAspect = (float) mMaxHeight * aspectRatio;
-                        if (widthForAspect < heightForAspect) {
-                            docWidth = Math.round(widthForAspect);
-                            docHeight = mMaxHeight;
-                        } else {
-                            docWidth = mMaxWidth;
-                            docHeight = Math.round(heightForAspect);
-                        }
+                    if (aspectRatio != -1) {
```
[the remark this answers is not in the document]
Fs00: Since the method explicitly returns -1 when the ratio cannot be determined, checking for that specific value looked more intention-revealing to me.
Does the safety concern apply here? I think it would if we did the check against the result of a floating-point operation, but getAspectRatio does a plain `return -1` inside. Am I missing something?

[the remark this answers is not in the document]
maniac103: I generally try to rely as little as possible on implementation details of 3rd party code, which could change at any time. The `> 0` check basically meant to express/test 'a valid aspect ratio was returned'. If the lib returned e.g. -0.5, we surely do not want to use that value.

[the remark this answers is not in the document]
Fs00: Makes sense. In this case it wasn't really an implementation detail since the behavior is explicitly documented in the JavaDoc, but I'm perfectly fine with reverting it if you prefer.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
Fs00: > I'm not sure what the point of that is?

Well, the point of scaling is that if you don't do it, an image wider than the screen will be rendered at full size and the parts that don't fit in the screen (or better, in the TextView) will be cut off.
I totally agree that this solution is far from ideal and that we should use the TextView size instead. The problem is that doing this would require a quite big refactor since the TextView is stored in `ObjectInfo` class... we should at the very least move the scaling logic there.

I've used 48dp since it's the highest TextView padding I could find in the app (the repository README TextView). It's not a big deal if we remove this for now, the only downside is that the border of some wide images might get cut off a bit.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
```diff
@@ -538,32 +535,25 @@ private Bitmap renderSvgToBitmap(Resources res, InputStream is) {
...
+                        // we make it smaller than maxWidth so that it doesn't take too much screen space
+                        docWidth = Math.round(mMaxWidth * 0.7f);
```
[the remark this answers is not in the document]
Fs00: Right, I should have explained it in the PR description 😅
The previous logic ended up setting `docWidth` to `mMaxWidth / 2` in these cases, because `docWidth` was initially set to `mMaxWidth` and then divided by 2 in the while loop below since `docWidth` wasn't `< mMaxWidth`.
The `mMaxWidth * 0.7` is an arbitrary choice I've made to leverage a bit more screen space than `mMaxWidth / 2`, and at the same time to avoid using an excessive amount of space (which would happen if we used `mMaxWidth` directly).

Also, if we revert the [above change](https://github.com/slapperwan/gh4a/pull/1230#discussion_r961362156), not using `mMaxWidth` as the `docWidth` becomes more important because it would prevent a bit of image border from being cut off.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
Fs00: You can see it in the app with the following image, whose root `<svg>` element has a width specified in `pt`s:
width_pt

Setting the render DPI to a different value causes the above image to be scaled incorrectly. More specifically, sizes returned by `getDocumentWidth/Height` will be way higher than the actual sizes of the image (because `DENSITY_DEFAULT` is higher than the default value of 96 used by the library) and therefore you'll see a big margin around the image because the `density` variable used a few lines below gets divided by 2 more times than it's actually necessary.
This change doesn't affect SVGs whose size is defined in pixels, because the render DPI is only used to convert "real world" units (pts, cms, inches) in pixels.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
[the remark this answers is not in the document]
Fs00: Reverted this. Let me know about the rest.

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
maniac103: The thing is, rendering at 96dpi does not make it correct, since the pixel density of the device's screen is not `96 dpi * density`, but `160 dpi * density`. The basic idea is
- make the library render the image at `mdpi` (160 dpi)
- do the upscaling for higher densities via canvas scaling

Where does the default 96 dpi come into play here?

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
Fs00: Reverted

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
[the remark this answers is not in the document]
Fs00: Reverted

app/src/main/java/com/gh4a/utils/HttpImageGetter.java
Fs00: For now I've reverted the change.
As it is now there is clearly some issue with the scaling, either with the DPI or with the canvas scaling. Removing the line makes the image look way more correct (just look at the comment above in the app) but I don't know if that's the most correct way to fix it.
By the way, currently the app does not scale raster images according to the screen DPI either.
My feeling is that we should use a library like Coil or Picasso to handle image loading and rendering instead of doing it manually... it's a burden and quite error prone.

Fs00: > can not be meaningfully scaled without quality loss

That's the price to pay when using high DPI displays ¯\_(ツ)_/¯... you can simply open the OctoDroid Readme from the Files list to see how the WebView properly scales raster images, compared to the Readme preview in the repository About tab.

Regarding libraries, [Coil](https://coil-kt.github.io/coil/) doesn't force you to load images in an ImageView: it can fetch and return them as Drawables after scaling/transforming them. It also has built-in support for SVGs backed by the same library we use.
Once minSdk is bumped to 21, it would be a viable way to simplify the code and avoid doing all the work manually.

PR #1244, opened 2022-10-09
Fix a few PR-related bugs

Fs00:
This PR batches together a few bug fixes, mostly regarding PRs (it's easier to review commit by commit):
- avoid showing the "View in file" action in review comments and diff fragments if the entire patch for that file wasn't returned by GH API (e.g. when the diff is too big). Previously, when tapping on "View in file" in these cases, the code viewer would open with a blank screen.
- completely hide (not just disable) Merge and Reopen actions in the app bar when a PR is already merged, since those actions are not possible in that context.
Also, the condition used to determine PR mergeability has been fixed: the Merge action will now be disabled when a PR can't be merged due to conflicts or missing approval.
- the app now shows the error snackbar when the deletion of a pull request branch fails and GH returns a 404 error (previously the error wasn't reported to the user)
- revamp (again) the merge dialog since a recent update of the Material components library broke the appearance of multi-line text field hints, as you can see below.
My solution was to get rid of the hints altogether, and use a helper text under the commit title field without "guessing" the default title (which is even better since users [can now change the default merge commit message](https://github.blog/changelog/2022-08-23-new-options-for-controlling-the-default-commit-message-when-merging-a-pull-request/)).
<table>
<tr><td>

</td><td>
</td></tr>
</table>

- small extra: POT files (gettext translation templates) were incorrectly recognized as binary files by the app and therefore were opened in the browser. That's been fixed too.

### DIFF
--- a/app/src/main/java/com/gh4a/activities/PullRequestActivity.java
+++ b/app/src/main/java/com/gh4a/activities/PullRequestActivity.java
@@ -22,23 +22,6 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.Bundle;
 import android.os.Parcelable;
-
-import androidx.activity.result.ActivityResultLauncher;
-import androidx.activity.result.contract.ActivityResultContracts;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.annotation.StringRes;
-
-import com.gh4a.utils.ActivityResultHelpers;
-import com.gh4a.utils.UiUtils;
-import com.google.android.material.appbar.AppBarLayout;
-
-import androidx.appcompat.view.ContextThemeWrapper;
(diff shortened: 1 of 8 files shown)


Fs00:
I have only a doubt on one thing @maniac103: since the mergeability check is now more accurate, when the user loads a PR and its mergeable state is yet to be determined by GitHub (unknown), the Merge button will be greyed out. If the PR then turns out to be mergeable, the next time the user refreshes the PR screen the Merge button will be clickable.
I recognize that this is not ideal for the user experience, since in some cases (which shouldn't be the majority though) the user can't go and merge a PR straight away but needs to refresh the screen. I think this is better than having the Merge button clickable when the PR can't be merged (e.g. due to conflicts), but let me know your thoughts...

PR #1278, opened 2023-04-08
Improvements and fixes to issue/PR list filtering

Fs00:
The primary goal of this PR is to allow searching by text and filtering by milestone/label etc. at the same time in issue/PR lists.
Previously, searching for issues by text would silently ignore the filters applied by the user, which was IMHO a confusing and unnecessarily restrictive behavior.
While implementing this, I've also discovered and fixed some issues:
- in the right drawer, the currently selected milestone/label/etc. wasn't being displayed because menu items' text was truncated to 1 line by default

| Before fix | After fix |
|--------|--------|
| Screenshot_filter_before | Screenshot_filter_after | 

- "Filter by X" dialogs would always start with the first item selected, no matter what the current selection was. The issue was caused by a mistyped Fragment argument name, so I've moved those names to constants to avoid the same problem in the future.
- the empty placeholder text (the one shown when there are no matching issues) now displays the correct message when the user is filtering but not searching by text (previously it showed "No issues have been reported so far")
- small tweak: I've renamed "Filter by labels" menu item into "Filter by label" because it allows searching only by a single label and not by multiple ones

### DIFF
--- a/app/src/main/java/com/gh4a/activities/IssueListActivity.java
+++ b/app/src/main/java/com/gh4a/activities/IssueListActivity.java
@@ -94,6 +94,7 @@ public class IssueListActivity extends BaseFragmentPagerActivity implements
     private int mSelectedParticipatingStatus = 0;
 
     private FloatingActionButton mCreateFab;
+    private MenuItem mRemoveFilterButton;
     private IssueListFragment mOpenFragment;
     private Boolean mIsCollaborator;
     private List<Label> mLabels;
@@ -399,6 +400,7 @@ public class IssueListActivity extends BaseFragmentPagerActivity implements
     @Override
     public boolean onCreateOptionsMenu(Menu menu) {
         getMenuInflater().inflate(R.menu.issue_list_menu, menu);
+        mRemoveFilterButton = menu.findItem(R.id.remove_filter);
 
         MenuItem searchItem = menu.findItem(R.id.search);
         searchItem.setOnActionExpandListener(this);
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/gh4a/activities/IssueListActivity.java
```diff
@@ -94,6 +94,7 @@ public static Intent makeIntent(Context context, String repoOwner, String repoNa
     private int mSelectedParticipatingStatus = 0;
 
     private FloatingActionButton mCreateFab;
+    private MenuItem removeFilterButton;
```
maniac103, line 97: `mRemoveFilterButton`

maniac103 (commented):
Other than the nitpick, LGTM :+1:

app/src/main/java/com/gh4a/activities/IssueListActivity.java
Fs00: Done.
I like this naming convention so much that I keep forgetting about it :joy:

maniac103: Nowadays, when writing Kotlin code, I omit the `m` prefix as well, but let's keep the code base consistent ;-)

Fs00: I agree I agree, it's just something very unnatural if you aren't used to it😅
Since you mentioned Kotlin, what about introducing it in the app? In addition to helping reduce boilerplate, it would allow us to use coroutines which are IMO way easier to approach than RxJava.

maniac103: > Since you mentioned Kotlin, what about introducing it in the app? In addition to helping reduce boilerplate, it would allow us to use coroutines which are IMO way easier to approach than RxJava.

As much as I love Kotlin, that would be a huge rewrite (since one would need to rewrite the bindings for that as well). I'm happy to make you the maintainer of both, but I'm afraid this is too much effort for my personal time budget. 😔

Fs00: I didn't mean to rewrite the whole app using Kotlin + coroutines, that would be really too much 😅
My idea is to at least make Kotlin available for developing new parts and for gradually migrating existing portions of the codebase to Kotlin as we work on them.
Regarding coroutines, we wouldn't need to rewrite the bindings to use them since the Kotlin coroutines library [offers some utilities for RxJava 2 interoperability](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-rx2/). I'm not sure on what could be the best way to integrate the two libraries (and if it's a good thing to do), but having Kotlin in the project would give us the opportunity to experiment.

PR #1281, opened 2023-04-21
Material Design 3 migration

### LINKED ISSUE #1016: Follow material design 2 guidelines.

### LINKED ISSUE #1215: Can we expect a new design
Sonething like modren (material ui 3 ) type or something?

### LINKED ISSUE #1245: Add themed icon (Android 13)
https://developer.android.com/about/versions/13/features#themed-app-icons

### LINKED ISSUE #1268: [Feature Request] redesign with MD3
MD1 is outdated. Maybe we should redesign the app with MD2 or MD3(you) ?

### LINKED ISSUE #1279: Add monochrome app icon for Android 13
For Android 13 apps can provide a monochrome launcher icon that can be used by launchers: https://developer.android.com/develop/ui/views/launch/icon_design_adaptive

Bnyro:
Hey, 
this is an attempt to migrate the app to [Material Design 3](https://m3.material.io).
I'm interested in other's opinions about it and open for some improvements :)

closes #1268 
closes #1279
closes #1245 
closes #1215
closes #1016


maniac103:
Got some screenshots? 😉

Bnyro:
Oh wow, you're very fast 😂
I'll upload some the next minutes :)

Bnyro:
Screenshot_20230421-153859_OctoDroid
Screenshot_20230421-153912_OctoDroid
Screenshot_20230421-153918_OctoDroid
Screenshot_20230421-154102_OctoDroid
Screenshot_20230421-154108_OctoDroid
Screenshot_20230421-154143_OctoDroid
Screenshot_20230421-154212_OctoDroid

Bnyro:
Note that the screenshots were made on Android 13. On Android 12 and up, wallpaper based colors are used, while on Android 11 and below a green color scheme based on the app's current theme will be used.

Bnyro:
And one thing I haven't been able to figure out is where the colors of the toolbar are getting changed inside the code, because in theory they should follow the normal color scheme and not have any issues with the visibility of the colors (currently, it's sometimes hard to see the toolbar text)

Bnyro:
And the settings:
Screenshot_20230421-155230_OctoDroid
Screenshot_20230421-155235_OctoDroid

maniac103:
I haven't specifically looked at the code, but from the screenshots, it mostly looks good visually, with one exception, which is the recoloring. TBH, I don't like it at all - IMO, when using it, the app needs to use the theme colors sparingly, which it currently doesn't do at all.
It would be great if you could
- use this PR for the restyling (dialogs, navigation drawer, all-caps titles etc.)
- move the misc. Android 13 fixes (monochrome icon, notification permissions) (thanks for those!) to a separate PR
- move out the color changes to a separate PR

Thanks :-)

Bnyro:
To which color changes are you referring to in particular? The dynamic wallpaper based colors on Android 12 and up, or the colors for opened/closed/draft issues/PRs? I just want to note that not using Material Design 3 colors is what I've done when starting to work on the PR and as per my opinion, it simply doesn't fit the style of Material Design 3 to use these red/green colors in combination with it. 

And sure, I'll make a separate PR for the icon + SDK bump in the next days ;)

Bnyro:
Dynamic (wallpaper based) colors could also become a preference only on Android 12 and up so that the green colors can still be used.

maniac103:
> To which color changes are you referring to in particular? The dynamic wallpaper based colors on Android 12 and up, or the colors for opened/closed/draft issues/PRs?

Both actually. As mentioned, I was judging from screenshots, not the code, so I didn't even notice there was more to it than the dynamic colors.

> as per my opinion, it simply doesn't fit the style of Material Design 3 to use these red/green colors in combination with it.

The thing is: the choice of both the green, red and purple have a background (the GitHub UI, and in case of the green it's also the icon). It's probably fine to replace them with dynamic colors, but IMHO that should happen *after* a full MD3 redesign, extending way beyond replacing the top level theme. As mentioned, we'd need to reduce color usage overall, probably replace tab bars by bottom navigation, probably restyle list contents etc. pp. When all of that is done and we have mostly white app contents with sparingly used color highlights, *then* (again, IMHO) it makes sense to rethink the color scheme.

Bnyro:
Okay, fair enough.
What'd be your thought on creating a new branch (something like md3-migration) which we'll use this PR as starting point for and then redesign some not well fitting components (the list items, the issue/PR activities, etc) from time to time. That way everyone interested could contribute to the migration and we wouldn't deliver a half finished migration as a release.

Unfortunately I'm maintaining some other projects currently which already require almost all my free time and attention, so I doubt I'd be able to do that all alone, though I'd make sure to work on it too when I find the time.

mueller-ma:
> move the misc. Android 13 fixes (monochrome icon, notification permissions) (thanks for those!) to a separate PR

Done in #1287

Bnyro:
As supposed, I've now
1. replaced the colored top app bars with Material Design 3 styled top app bar (with only a very low accent)
2. removed all code that changes the header, tab and statusbar colors
3. Redesigned the issues and PR header
4. Slightly migrated the comment editor to MD3
5. slightly changed the appearance of some list items

I think that the other list items that weren't touch yet are fine to stay like they are for now as they go well with the style changes, apart from that `ListItem` components are not available in MDC Android but only in Jetpack Compose, hence re-designing all of them requires some creativity as well.
I hope someone can test the changes in the next days, I've been running the branch for a week now and fixed all the issues I've found.

Fs00:
Hopefully I'll give it a whirl this weekend, thanks for this!
At the moment I have a question: why has the SmoothProgressBar been replaced with the MDC one? IMHO the former looks much nicer and and also gives a distinct look to the app.

Bnyro:
I've been removing it because it has been using different colours based on the issue/PR state before, and as these colours were removed I thought it'd be reasonable to also switch to using the material design progress indicator as it's only based on one color (the primary one).
However, we could also use the old/smooth one and use the primary and secondary colours, however I'm not sure if this would fit the style of Material Design 3 too well.
So I guess we should continue discussing it if you don't feel comfortable with the material progress indicator when testing it :)

Fs00:
Hello @Bnyro, I've finally found the time to test this PR and write down my findings.
IMO there's still some work to do, but the redesign is already looking very promising. Here's my feedback on it (note that I've tested it on Android 10 using the light theme most of the time):
- in light theme, the background color of the content area isn't always consistent: sometimes it's white, sometimes it's light green. IMO we should go with the white background everywhere, so that the app bar has a bit of elevation compared to the content.
- while cards in dark theme look way better in this PR, the fact that they are both filled and elevated makes them stand out too much in light theme. I would just keep the elevation shadow and remove the fill color in the latter theme.
- I find the redesigned issue/PR header to be a substantial downgrade compared to the current one... while I appreciate the font change, the title text is truncated and not selectable/copyable by the user anymore.
Also, being on its own line, the state indicator uses too much vertical space (considering that the header is already quite tall, especially in landscape mode) and the shadow behind the "Open"/"Closed" text looks broken in light theme.
- I really like the refreshed comment editor! A small suggestion: what about replacing the collapse icon with a flipped version of the expand icon?
- I'm still of the opinion that replacing the SmoothProgressBar makes the app lose personality, but I have to admit that the MDC one isn't as bad as I initially thought
- I'm not a fan of the new scrollbar color, which now appears to be 100% black or white depending on the theme (is it intentional though? I can't find the relevant change in the diff)

I've also found some bugs/oversights:
- all icons in the app bar (except for the overflow menu one) are always white, which is incorrect for the light theme
- in the issue edit screen, the title text field has a wrong color in both light and dark themes
- the app bar color flickers on scroll
- the background color of the welcome screen (the one shown when the user is logged out) hasn't been adjusted yet, since it's still the same as before (which looks out of place after the redesign)
- FAB colors are sometimes inconsistent: for example, in the issues list it's dark green, but the edit button in the issue screen is light green
- the comment editor's resizing and scrolling behavior in the collapsed state is now broken: try pasting this comment in it and you'll see it by yourself 😉

Keep up the good work!
PS: obviously @maniac103 has the last word on the design of the app, so his opinion will be definitely more important than mine 🙂

Bnyro:
Thanks for the review, I added some comments about some things and will track my progress with working on it below:

>     * in light theme, the background color of the content area isn't always consistent: sometimes it's white, sometimes it's light green. IMO we should go with the white background everywhere, so that the app bar has a bit of elevation compared to the content.
> 
TODO
>     * while cards in dark theme look way better in this PR, the fact that they are both filled and elevated makes them stand out too much in light theme. I would just keep the elevation shadow and remove the fill color in the latter theme.
> 
TODO
>     * I find the redesigned issue/PR header to be a substantial downgrade compared to the current one... while I appreciate the font change, the title text is truncated and not selectable/copyable by the user anymore. Also, being on its own line, the state indicator uses too much vertical space (considering that the header is already quite tall, especially in landscape mode) and the shadow behind the "Open"/"Closed" text looks broken in light theme.
> 
I haven't tested it with the light theme earlier, I fixed the appearance there now. Though I don't agree with your other concerns, in my opinion the issue state indicator doesn't require too much space, just looks right in my eyes (very subjective though). And the text is still selectable though, it's just not very convenient.
>     * I really like the refreshed comment editor! A small suggestion: what about replacing the collapse icon with a flipped version of the expand icon?
> 
That's really a good idea, applied your suggestion!
>     * I'm still of the opinion that replacing the SmoothProgressBar makes the app lose personality, but I have to admit that the MDC one isn't as bad as I initially thought
> 
The SmoothProgressBar library didn't get updated in a while and I think it's better to just use Material Design 3 components whenever possible.
>     * I'm not a fan of the new scrollbar color, which now appears to be 100% black or white depending on the theme (is it intentional though? I can't find the relevant change in the diff)
> 
Yes, that seems to be a bit buggy, I didn't intend to change it like that. Improved that now!

>     * all icons in the app bar (except for the overflow menu one) are always white, which is incorrect for the light theme
> 
TODO
>     * in the issue edit screen, the title text field has a wrong color in both light and dark themes
> 
Thanks for the catch, fixed.
>     * the app bar color flickers on scroll
> 
That's intended, it got introduced with Material Design 3 and is enabled by default.
>     * the background color of the welcome screen (the one shown when the user is logged out) hasn't been adjusted yet, since it's still the same as before (which looks out of place after the redesign)
> 
Fixed now!
>     * FAB colors are sometimes inconsistent: for example, in the issues list it's dark green, but the edit button in the issue screen is light green
> 
True, I tried to make it look consistent now, at least any FloatingActionButton I was able to find looks the same now.
>     * the comment editor's resizing and scrolling behavior in the collapsed state is now broken: try pasting this comment in it and you'll see it by yourself
> 
TODO: Yes, I see the issue, help with that would be appreciated, I don't really know how to fix that properly.

Secret-chest:
NOOOOO!

I hate Material You. I don't know why, but it makes me dizzy, it is too cramped, rounded and padded, the background is not pure and everything is pastel.

Remember this is a developer app.

Bnyro:
Unfortunately, I no longer use the app and thus don't have any interest in continuing this PR.

If anyone wants to, feel free to use my changes and continue working on this on a new PR!

guguss-31:
Any updates ?

PR #1282, opened 2023-04-21
Make launcher icon independent from DayNight mode

mueller-ma:
Before the launcher icon background was brighter when the device was in day mode when the icon has been added. It didn't follow the DayNight mode, so it felt wrong to me to respect DayNight mode.

### DIFF
--- a/app/src/main/res/drawable-anydpi-v26/launcher_bg.xml
+++ b/app/src/main/res/drawable-anydpi-v26/launcher_bg.xml
@@ -9,8 +9,8 @@
         android:pathData="M0,0 H100 V100 H-100 Z">
         <aapt:attr name="android:fillColor">
             <gradient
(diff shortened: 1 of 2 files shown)


Fs00:
@maniac103 @mueller-ma I've noticed the color change while trying some debug builds lately and I was thinking that probably the app icon should always use the brighter shade of green, shouldn't it? Especially given that the latter is based on the light theme palette, which is the default one on all Android <10 devices, and the dark theme is derived from it.
It would have been cool if icon theming worked though...

maniac103:
Yes, agreed, I should have paid more attention here.

mueller-ma:
On my device the dark mode is the "default" theme, so I didn't think about using the bright color. I opened https://github.com/slapperwan/gh4a/pull/1286 to use the brighter color again.

PR #1286, opened 2023-04-25
Use a brighter icon background

mueller-ma:
See https://github.com/slapperwan/gh4a/pull/1282#issuecomment-1521930094


maniac103:
Thanks :+1:

PR #1290, opened 2023-04-29
Make WebView stylesheets more maintainable and consistent

### LINKED ISSUE #1274: Hightlighted source code contrast is illegible in night mode
It is almost impossible to read highlighted source code in night mode.

Example link: https://github.com/linux-surface/iptsd/blob/master/etc/iptsd.conf#L62-L78 or more permanently https://github.com/linux-surface/iptsd/blob/3ec9be098bd42d8fa1f7f49c13689b174edb30cb/etc/iptsd.conf#L62-L78

# Expected Behavior

Text contrast should follow accessibility guidelines. It should be white when the highlight is dark yellow.

# Actual Behavior

Grey text on dark yellow highlight is unreadable:

Screenshot_20230226-060329_OctoDroid

Fs00:
The primary goal of this PR is to leverage the `@import` feature in CSS (available from API 21 WebView) to greatly reduce duplication in stylesheets used in WebViews.
With these changes, the _dark_ and _print_ variants of the stylesheets will derive from the _light_ variant using `@import`, and then override only the parts that differ. This will make maintainance a lot easier and helps keep the styles consistent: in fact, before this PR prettify (code preview) stylesheets had several differences between dark and light themes, but now no more:

| Before | After |
|--------|--------|
| Screenshot_20230429-114651_OctoDroid | Screenshot_20230429-114434_OctoDroid_Debug | 

While I was at it, I've also taken the opportunity to fix #1274 ([here's how repro looks like after the fix](https://user-images.githubusercontent.com/30041551/235297000-894176a1-264f-4858-b556-8eee880921d5.png)) and tweak code background colors in Markdown CSS to make it more similar to the one rendered in TextViews.

Last but not least, I've found a neat CSS trick to make Markdown tables horizontally scrollable and prevent them from overflowing.


app/src/main/assets/prettify-print.css
```diff
@@ -1,80 +1,13 @@
...
 
-.tag {
-    color: #008;
+pre.prettyprint.linenums {
+    box-shadow: none;
 }
 
-.atn {
-    color: teal;
+li.L1,li.L3,li.L5,li.L7,li.L9 {
```
Bnyro, line 7: This could also be something like 
```css
li:nth-child(even) {
    background: none;
}
```
maybe, as we want every second line to look different (should be supported by WebView in theory, not tested yet.)

Fs00: Initially I just used the same rule as the other stylesheets, but this selector works the same ([right from API 21 WebView](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child#browser_compatibility)) and also looks clearer :)

PR #1298, opened 2023-05-29
Use dynamic status bar height in the left drawer header

raphaelbussa:
Use dynamic status bar height in the left drawer header instead of the 24dp hard coded

Before / After

Screenshot_20230529_190130 MediumScreenshot_20230529_190100 Medium

### DIFF
--- a/app/src/main/java/com/gh4a/BaseActivity.java
+++ b/app/src/main/java/com/gh4a/BaseActivity.java
@@ -51,6 +51,8 @@ import com.google.android.material.snackbar.Snackbar;
 import androidx.core.app.ActivityCompat;
 import androidx.core.content.ContextCompat;
 import androidx.core.view.GravityCompat;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.WindowInsetsCompat;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/gh4a/BaseActivity.java
```diff
@@ -275,6 +277,11 @@ protected void toggleRightSideDrawer() {
     }
 
     protected void configureLeftDrawerHeader(View header) {
+        ViewCompat.setOnApplyWindowInsetsListener(header, (view, insets) -> {
+            int statusBarHeight = insets.getInsets(WindowInsetsCompat.Type.systemBars()).top;
+            view.setPadding(0, statusBarHeight, 0, 0);
```
maniac103, line 282: Won't this kill the left and bottom padding set in XML?

raphaelbussa: Yes, but the root layout has no padding -> https://github.com/slapperwan/gh4a/blob/6080d0e6ada126ce70654b7fdd3f319bbff83e9a/app/src/main/res/layout/drawer_header_left.xml#L8

PR #1345, opened 2024-07-06
Re-introduce ability to install downloaded APKs directly

Fs00:
Since we no longer target the Play Store, we can re-add the permission that allows downloaded APKs to be installed by tapping directly on the download notification. This was initially introduced in #1130 and reverted in f6f325e.


Fs00:
@maniac103 No hurry, but can this PR and #1290 be merged before the next release?

PR #1353, opened 2024-07-07
Replace custom ghFont attribute with native Android attributes

Fs00:
This PR applies the remaining changes that were initially introduced in #1166, but were then discarded to make the PR easier to review.
This finally gets rid of all the infrastructure that was required to use custom typefaces on API <21.

Compared to #1166, I've added a couple of minor changes:
- removed lookup of the `needsLinkHandling` attribute in TextAppearance styles, which only made sense for `ghFont` since the latter was used in a couple of TextAppearance styles
- enabling `autoLink` on a StyleableTextView now enables the behavior provided by `needsLinkHandling=true`. This makes the behavior of the TextView more intuitive and prevents non-functional links caused by not setting `needsLinkHandling` (as it already happened in the past)

I also think that we should find a new name for `StyleableTextView`. I guess that this name was chosen because the TextView could be _styled_ via the `ghFont` attribute? Since the main feature it provides now is link handling, we could call it `LinkHandlingTextView`, but let me know your thoughts...


maniac103:
> I also think that we should find a new name for StyleableTextView. I guess that this name was chosen because the TextView could be styled via the ghFont attribute? Since the main feature it provides now is link handling, we could call it LinkHandlingTextView, but let me know your thoughts...

:+1: to both the explanation of existing name and naming suggestion.

Fs00:
Renaming `StyleableTextView` into `LinkHandingTextView` made me realize that we don't need to use it in most cases. Regular TextViews are perfectly fine when we don't have any links or text is not selectable.
Because of that, I've pushed a commit to remove references to `LinkHandlingTextView` in Java code when TextView methods alone are used, and another one to replace `LinkHandlingTextView`s with TextViews in several XML layouts (mostly clickable rows used in RecyclerViews).
I've double checked and tested to make sure that the behavior of the UI stayed the same. Given the large number of touched files, it would be best if this PR didn't sit here for too long :pray:

maniac103:
Thanks!

PR #1373, opened 2024-07-24
Add ability to select a reason when closing an issue

Fs00:
This PR implements the ability to specify a reason (_completed_ or _not planned_) when closing an issue by tapping the "Close" action in the app bar.
In #1223 I initially thought that this couldn't be done with the REST API, but luckily I was wrong.

close_as_new

I intentionally kept the close reason descriptions short (almost the same as the GH web UI) so that they fit on a single line even on smaller screens. 

One minor feature that this PR lacks compared to the GH web UI is the ability to change the close reason of an issue without first reopening it. I didn't bother to add it, since I think it will rarely be useful in practice.
In case we really want to have it, we could simply add a "Close as not planned"/"Close as completed" action in the dropdown menu when the issue has already been closed.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -432,8 +432,13 @@
     <string name="issue_comment_hint">Comment on this issue</string>
     <string name="commit_comment_header">%1$s added a note %2$s</string>
     <string name="issue_no_description">No description provided.</string>
-    <string name="close_issue_confirm">Do you really want to close this issue?</string>
     <string name="reopen_issue_confirm">Do you really want to reopen this issue?</string>
+    <string name="close_issue_dialog_title">Close issue as\u2026</string>
+    <string name="issue_reason_not_planned">Not planned</string>
+    <string name="issue_reason_not_planned_description">Won\'t fix, duplicate, stale</string>
+    <string name="issue_reason_completed">Completed</string>
+    <string name="issue_reason_completed_description">Done, closed, fixed, resolved</string>
+    <string name="close_issue">Close issue</string>
     <string name="issue_template_dialog_title">Select issue type</string>
 
     <!-- Pull Request -->
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -49,6 +49,7 @@ android {
     }
 
     compileOptions {
+        coreLibraryDesugaringEnabled true
         sourceCompatibility JavaVersion.VERSION_17
         targetCompatibility JavaVersion.VERSION_17
     }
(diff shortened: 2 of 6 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -432,8 +432,13 @@
     <string name="issue_comment_hint">Comment on this issue</string>
     <string name="commit_comment_header">%1$s added a note %2$s</string>
     <string name="issue_no_description">No description provided.</string>
-    <string name="close_issue_confirm">Do you really want to close this issue?</string>
     <string name="reopen_issue_confirm">Do you really want to reopen this issue?</string>
+    <string name="choose_reason">Choose a reason</string>
```
maniac103, line 436: How about `Close issue as\u2026`?

Fs00: I agree that it suits better. I've pushed the change and updated the screenshot in the description.

maniac103:
Thinking about it, maybe there's potential to consolidate with [this code](https://github.com/slapperwan/gh4a/blob/master/app/src/main/java/com/gh4a/activities/IssueEditActivity.java#L827) used in a similar dialog? It uses a dedicated adapter instead of the (creative, but still ;-) ) string builder hack.
Maybe refactor into a `TwoLineSelectionDialog` or something? Or share the adapter at least?

Fs00:
Hmm... I think that having a single dialog for both use-cases will not help, since the user interaction is different between the two (checking a radio button and confirming _vs_ just tapping on an item in the list).
Also, having a common adapter doesn't seem worth it because the item layout is different in this case, since we have the radio button in addition to just the title and description.
If you'd like to avoid the code messing around with the `SpannableStringBuilder`, I could try to use a dedicated adapter + layout for the items (something similar to the solutions proposed in [this SO thread](https://stackoverflow.com/questions/59696021/how-to-add-title-and-subtitle-in-android-radiobutton)), although I think that the implementation will be a bit more complex overall.

maniac103:
> Hmm... I think that having a single dialog for both use-cases will not help, since the user interaction is different between the two (checking a radio button and confirming vs just tapping on an item in the list).

Why not make both confirm on item tap? ;-) Technically, the current solution is 'confirm on item tap' as well, even though the item currently is a dialog button.

Fs00:
Could be an interesting idea, I'll try and see how it looks

Fs00:
I've just updated the implementation to use the same layout and behavior as the issue template selection dialog (screenshot in the description updated), which allowed me to reuse the same adapter for both dialogs.
While I was at it, I've discovered and fixed two minor bugs in the issue template dialog:
- the order of issue templates as returned by the API was not deterministic, so we now sort them by name
- tapping the Cancel button didn't fire the `onCancel` callback as the previous code assumed, and therefore in that case the editor would have remained stuck on the "Loading issue templates" message

To make my life easier, I have also enabled support for Java 17 features (which include records, switch expressions, text blocks and some more) and desugaring for Java 8/11 library features (for using streams and `Comparator.comparing`).

PR #1402, opened 2024-09-29
Add support for release reactions

Fs00:
This PR implements full support for viewing and toggling reactions on releases. Here's an example:

reactions

The implementation has been harder than I initially hoped for a few reasons:
- GitHub allows to react on releases only with "positive" reactions (so no :-1: or :confused:), and that's where I started going down the rabbit hole.
The approach I chose - which to me looked the less impactful for existing logic in the ReactionBar - was to use a different menu resource for the "add reaction popup" without :-1:/:confused: items, and set up the ReactionBar use that.
However, the `AddReactionMenuHelper` wasn't designed to support showing only a subset of all possible reaction types in the menu. So I had to heavily refactor it to accommodate for that, replacing parallel arrays used to store menu items and user's own reaction IDs with a `List` and a `HashMap` respectively. I've used this opportunity to simplify the logic of several methods of the class (using Streams where possible) and improve some method/field names (commits 0a7f49d and 3bf3077)
- when the user is logged into the app via OAuth, API calls for reacting on a release always fail if the repo belongs to an org that has [OAuth restrictions enabled](https://docs.github.com/en/organizations/managing-oauth-access-to-your-organizations-data/about-oauth-app-access-restrictions) (two notable examples I've encountered are PowerShell and rails).
While it looks like a weird API behavior (or maybe a bug, I'll ask GH support when I have some time), it's also true that the app didn't show an error message if toggling a reaction failed. So I've decided to handle this by displaying a snackbar as we already do in other situations when a user action fails. I plan to try to improve those snackbars with more details on the error in a future PR.

I recommend reviewing this PR commit by commit. I have added some review comments to explain the context behind some smaller changes.

### DIFF
--- a/app/src/main/java/com/gh4a/widget/ReactionBar.java
+++ b/app/src/main/java/com/gh4a/widget/ReactionBar.java
@@ -7,6 +7,7 @@ import android.os.Parcelable;
 import androidx.annotation.ColorInt;
 import androidx.annotation.IdRes;
 import androidx.annotation.LayoutRes;
+import androidx.annotation.MenuRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.graphics.drawable.DrawableCompat;
@@ -85,14 +86,8 @@ public class ReactionBar extends HorizontalScrollView implements View.OnClickLis
 
     private ReactionDetailsCache mDetailsCache;
     private MenuPopupHelper mAddReactionPopup;
-    private AddReactionMenuHelper mAddHelper;
-    private final PopupMenu.OnMenuItemClickListener mAddReactionClickListener =
-            new PopupMenu.OnMenuItemClickListener() {
-        @Override
-        public boolean onMenuItemClick(MenuItem item) {
-            return mAddHelper.onItemClick(item);
-        }
-    };
+    private @MenuRes int mAddReactionMenuResId = R.menu.reaction_menu;
+    private AddReactionMenuHelper mAddReactionMenuHelper;
 
     public ReactionBar(Context context) {
         this(context, null);
@@ -124,8 +119,8 @@ public class ReactionBar extends HorizontalScrollView implements View.OnClickLis
         if (mPopup != null) {
             mPopup.update();
         }
-        if (mAddHelper != null) {
-            mAddHelper.updateMenuItems();
+        if (mAddReactionMenuHelper != null) {
+            mAddReactionMenuHelper.updateMenuItems();
         }
         if (reactions != null && reactions.totalCount() > 0) {
             updateView(mPlusOneView, reactions.plusOne());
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/gh4a/fragment/ReleaseListFragment.java
```diff
@@ -46,9 +46,7 @@ protected Single<Response<Page<Release>>> loadPage(int page, boolean bypassCache
 
     @Override
     protected RootAdapter<Release, ? extends RecyclerView.ViewHolder> onCreateAdapter() {
-        ReleaseAdapter adapter = new ReleaseAdapter(getActivity());
-        adapter.setOnItemClickListener(this);
```
Fs00, line 50: I've removed the call to `setOnItemClickListener` because it is already performed by the caller of `onCreateAdapter`: https://github.com/slapperwan/gh4a/blob/43a0a01ed900a69d898ac83c56bfc44d33f8e2ee/app/src/main/java/com/gh4a/fragment/PagedDataBaseFragment.java#L84-L88

app/src/main/java/com/gh4a/widget/ReactionBar.java
```diff
@@ -266,7 +274,7 @@ public void show(String content) {
         }
 
         public void toggleOwnReaction(Reaction currentReaction) {
-            final long id = currentReaction != null ? currentReaction.id() : 0;
+            Long id = currentReaction != null ? currentReaction.id() : null;
```
Fs00, line 277: Not a required change, but I thought that using a nullable `Long` here (and in the `toggleReaction` method) could make it a bit clearer that the reaction ID may or may not be present.

app/src/main/java/com/gh4a/widget/ReactionBar.java
```diff
@@ -474,110 +481,99 @@ public AddReactionMenuHelper(@NonNull Context context, Menu menu,
...
+            String ownLogin = Gh4Application.get().getAuthLogin();
+            reactions.stream()
+                    .filter(reaction -> ApiHelpers.loginEquals(reaction.user(), ownLogin))
+                    .forEach(reaction -> mUserOwnReactions.put(reaction.content(), reaction.id()));
+            setReactionMenuItemsVisible(true);
         }
 
         public void startLoadingIfNeeded() {
-            if (mOldReactionIds != null) {
-                syncCheckStates();
```
Fs00, line 533: It turned out that this check here was not needed at all, I'm not even sure why it was added (maybe to avoid re-running the entire logic to update menu items in some situations?)
I believe that it could have also caused the popup menu to display incorrect data when `AddReactionMenuHelper` is used in `CommentViewHolder`, since the same instance of the view holder and the menu helper is reused between different comments and therefore the data stored in `mOldReactionIds` could have referred to a different comment than the one bound to the view holder at a certain moment.

Fs00:
Rebased on latest master. The diff is now smaller and thus easier to review, although I recommend doing it commit by commit due to the scope of the changes.

PR #1420, opened 2024-11-15
Build adjustments and minor dependency updates

### LINKED ISSUE #1319: App is missing on Google Play Store

### LINKED ISSUE #1422: App is not on Playstore
is this app not availible Google playstore anymore?
even when i click on the link in the Readme.md.

thanks,  i used this app several years.

Fs00:
Aside from some minor dependency updates (including GH bindings), this PR introduces some tweaks to the build configuration:
- removed the [dependency info block](https://developer.android.com/build/dependencies#dependency-info-play) from the APK, which is useless for apps that aren't published on the Play Store
- migrated to [non-transitive R classes](https://developer.android.com/build/optimize-your-build#use-non-transitive-r-classes), which are now the default for new projects
- renamed the signing config for release builds to `release` (was previously named `playStore`)
- added the configuration to [remove non-English strings from built APKs via `resConfigs`](https://developer.android.com/build/shrink-code#unused-alt-resources) (will need to be updated in case a new translation is added to the app). In my testing this saves ~700KB from the release APK, due to several translated strings coming from AppCompat and Material libraries which are hardly ever visible to the user. Note that this doesn't affect PrettyTime translations because they do not use Android resources.

I have also removed the outdated Play Store badge from the Readme (closes #1319, closes #1422) and extracted a one-line bugfix that was initially included in #1402 (0acc437, fixes a incorrect UI update that occurs when toggling commit comment reactions) since I ended up changing the same file in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This application provides access to [GitHub](https://github.com/) and lets you s
 
 Download
 --------
-<a href='https://play.google.com/store/apps/details?id=com.gh4a'><img alt='Get it on Google Play' src='https://play.google.com/intl/en_us/badges/images/generic/en_badge_web_generic.png' height="80px" align="left"/></a>[<img src="https://f-droid.org/badge/get-it-on.png" alt="Get it on F-Droid" height="80px">](https://f-droid.org/packages/com.gh4a/)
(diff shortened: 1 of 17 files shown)


Fs00:
Rebased to fix a merge conflict.
@maniac103 When you have some time, could you please look into this? Updating the dependencies would unblock some other improvements I'd like to make.

Fs00:
Thanks!

PR #1424, opened 2024-11-23
Simplify releases URL handling

Fs00:
This change has been extracted from #1402 since it can be merged on its own, even though it's a pre-requisite to make release reactions work properly.

It essentially changes the `ReleaseActivity` to always fetch the release from GH API on its own, instead of receiving it via intent extras.
This was mainly done to make sure that opening a release from the list always shows up-to-date reaction counters, but it also allows to simplify releases link handling by removing the intermediate `ReleaseLoadTask` that was previously used to fetch a release by tag and pass it to the `ReleaseActivity` via intent extras.

This PR also includes a commit to fix URL handling for some GitHub website pages (/edu, /education, /newsroom, /premium-support and /watching).


app/src/main/java/com/gh4a/fragment/ReleaseListFragment.java
```diff
@@ -46,9 +46,7 @@ protected Single<Response<Page<Release>>> loadPage(int page, boolean bypassCache
 
     @Override
     protected RootAdapter<Release, ? extends RecyclerView.ViewHolder> onCreateAdapter() {
-        ReleaseAdapter adapter = new ReleaseAdapter(getActivity());
-        adapter.setOnItemClickListener(this);
-        return adapter;
+        return new ReleaseAdapter(getActivity());
```
Fs00, line 49: The call to `setOnItemClickListener` was unnecessary because it is already performed by the caller of `onCreateAdapter`: https://github.com/slapperwan/gh4a/blob/43a0a01ed900a69d898ac83c56bfc44d33f8e2ee/app/src/main/java/com/gh4a/fragment/PagedDataBaseFragment.java#L84-L88

PR #1432, opened 2024-12-26
Add "Discussions" shortcut to repository screen

Fs00:
This PR adds a "Discussions" shortcut to the "Other information" section for repositories that have discussions enabled, which opens the Discussions page in a custom tab or browser.

other_info_discussions

I have decided not to put it near issues/pulls/forks etc. because it would have looked out of place, considering that there is no way to get the discussions count from the REST API.
While working on the code, I did some minor clean-ups to leverage lambdas/method references and removed some leftover dead code.

Open question: is _"Other information"_ still a good title for this section? Releases and discussions are not really "information", maybe just _"Other"_ would be a better fit?

### DIFF
--- a/app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
+++ b/app/src/main/java/com/gh4a/fragment/RepositoryFragment.java
@@ -18,6 +18,7 @@ package com.gh4a.fragment;
 import android.content.Context;
 import android.content.Intent;
 import android.graphics.Typeface;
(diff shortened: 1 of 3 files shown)


maniac103:
> is "Other information" still a good title for this section? Releases and discussions are not really "information", maybe just "Other" would be a better fit?

Maybe it's just me, but I find both to fit equally well. Releases maybe could be put alongside issues and PRs, though...

Issue #1332, opened 2024-06-29
Look like json numbers are being parsed as Int32. They appear to require at least a int64


uxmal:
### Error stack trace
```
com.squareup.moshi.JsonDataException: com.squareup.moshi.JsonDataException: Expected an int but was 2149728112 at path $.items[5].payload.comment.pull_request_review_id at $.items[5].payload.comment.pull_request_review_id
	at com.squareup.moshi.AdapterMethodsFactory$1.fromJson(AdapterMethodsFactory.java:101)
	at com.squareup.moshi.CollectionJsonAdapter.fromJson(CollectionJsonAdapter.java:81)
	at com.squareup.moshi.CollectionJsonAdapter$2.fromJson(CollectionJsonAdapter.java:55)
	at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:41)
	at com.squareup.moshi.internal.NonNullJsonAdapter.fromJson(NonNullJsonAdapter.java:43)
	at com.meisolsson.githubsdk.model.AutoValue_Page$MoshiJsonAdapter.fromJson(AutoValue_Page.java:61)
	at com.meisolsson.githubsdk.model.AutoValue_Page$MoshiJsonAdapter.fromJson(AutoValue_Page.java:23)
	at com.squareup.moshi.internal.NullSafeJsonAdapter.fromJson(NullSafeJsonAdapter.java:41)
	at retrofit2.converter.moshi.MoshiResponseBodyConverter.convert(MoshiResponseBodyConverter.java:45)
	at retrofit2.converter.moshi.MoshiResponseBodyConverter.convert(MoshiResponseBodyConverter.java:27)
	at retrofit2.OkHttpCall.parseResponse(OkHttpCall.java:225)
	at retrofit2.OkHttpCall.execute(OkHttpCall.java:188)
	at retrofit2.adapter.rxjava2.CallExecuteObservable.subscribeActual(CallExecuteObservable.java:45)
	at io.reactivex.Observable.subscribe(Observable.java:12284)
	at io.reactivex.internal.operators.observable.ObservableSingleSingle.subscribeActual(ObservableSingleSingle.java:35)
	at io.reactivex.Single.subscribe(Single.java:3666)
	at com.tspoon.traceur.SingleOnAssembly.subscribeActual(SingleOnAssembly.java:43)
	at io.reactivex.Single.subscribe(Single.java:3666)
[...]

PR #1433, opened 2025-01-01
Display more helpful error messages in snackbars

Fs00:
This PR adds more information to error snackbars when the GitHub API returns an error upon performing an action, in order to help users understand the reason without digging into the logcat:

In these cases, the "Retry" button - which is almost always useless for API errors - is replaced by "Copy error message", which should make it easy to get the full message when reporting an issue or when it's too long to fit in the snackbar, as it's the case with OAuth restriction errors:

When an action fails due to a non-API error, the usual "Retry" button is displayed instead, but the message has been reworded to make it clear that it's most likely a connectivity issue (which should be the only other likely cause of errors, unless we have a bug in the code).

Let me know your thoughts!

### DIFF
--- a/app/src/main/java/com/gh4a/utils/RxUtils.java
+++ b/app/src/main/java/com/gh4a/utils/RxUtils.java
@@ -1,10 +1,13 @@
 package com.gh4a.utils;
 
-import android.annotation.SuppressLint;
 import android.app.Dialog;
+import android.content.Context;
 import android.os.Bundle;
+import android.util.Log;
+
 import androidx.annotation.StringRes;
 import androidx.coordinatorlayout.widget.CoordinatorLayout;
+
 import com.google.android.material.snackbar.Snackbar;
 import androidx.fragment.app.DialogFragment;
 import androidx.fragment.app.FragmentActivity;
@@ -17,6 +20,7 @@ import com.meisolsson.githubsdk.model.SearchPage;
 
 import org.reactivestreams.Publisher;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
@@ -117,7 +121,7 @@ public class RxUtils {
         return upstream -> upstream
                 .compose(RxUtils::doInBackground)
                 .compose(wrapWithProgressDialog(activity, dialogMessageResId))
-                .compose(wrapWithRetrySnackbar(rootLayout, errorMessage));
+                .compose(wrapWithErrorSnackbar(activity, rootLayout, errorMessage));
     }
 
     public static <T> SingleTransformer<T, T> wrapWithProgressDialog(final FragmentActivity activity,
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/gh4a/utils/RxUtils.java
```diff
@@ -175,8 +178,17 @@ public void onDismissed(Snackbar snackbar, int event) {
...
-                        .show();
+                        .setTextMaxLines(3)
+                        .setAction(R.string.retry, view -> retryProcessor.onNext(0));
+
+                if (isGitHubApiError) {
+                    snackbar
+                        .setDuration(Snackbar.LENGTH_INDEFINITE)
+                        .setAction(
+                            R.string.snackbar_copy_error_action,
+                            view -> IntentUtils.copyToClipboard(context, "GitHub error", error.getMessage()));
```
maniac103, line 189: Nit: 
```suggestion
                            view -> IntentUtils.copyToClipboard(context, "GitHub error", error.getMessage())
                        );
```

app/src/main/java/com/gh4a/utils/RxUtils.java
```diff
@@ -175,8 +178,17 @@ public void onDismissed(Snackbar snackbar, int event) {
                                 }
                             }
                         })
-                        .setAction(R.string.retry, view -> retryProcessor.onNext(0))
-                        .show();
+                        .setTextMaxLines(3)
+                        .setAction(R.string.retry, view -> retryProcessor.onNext(0));
```
maniac103, line 182: I'd prefer this to happen in an else branch below; it's a little confusing right now with action being set here and overwritten below.

app/src/main/res/values/strings.xml
```diff
@@ -68,6 +68,9 @@
     <string name="load_failure_explanation_dmca">This repository was taken down for legal reasons.\nPlease see %1$s for details.</string>
     <string name="load_failure_explanation_parsing">There was an error while parsing GitHub\'s response. You can use the button below to report this issue to the app developers.</string>
     <string name="load_failure_explanation_with_reason">Loading the data from GitHub failed. GitHub told us the following reason:\n%1$s</string>
+    <string name="snackbar_generic_error">%1$s. Please check your Internet connection.</string>
```
maniac103, line 71: I'm not sure about the 'please check your connection' text. It's probably helpful for the first time, but probably also redundant (to annoying) afterwards.
Also, why capitalize 'internet'?

Fs00: > Also, why capitalize 'internet'?

That's likely me being used to writing it using the Italian spelling :sweat_smile: 

Maybe something along the lines of _"This was likely caused by a connection error."_ would fit better?

maniac103: I think I prefer the original string (`Please check your internet connection.`) to that.
To avoid showing it unnecessarily, how about only showing that specific message in case of `IOException` being thrown, and omitting it in case of other exceptions?

Fs00: I think that's a good idea. I'm not a big of fan of the "X failed" message without any explanation though, even if it should rarely occur in practice. In those cases, maybe we could replace the "Copy error message" with a button to copy the stack trace and encourage the user to report the issue?

maniac103: That might be a good idea, or might cause similar spam as in #1332 ... not sure yet ;-)

maniac103: Agreed, I think that's OK.

Fs00: I have rebased and pushed the requested changes and added the ability to copy the error stack trace in case of application errors. I can confirm it works as expected because I've caught a JSON exception due to a GH error response not matching the `ClientErrorResponse` class in the bindings.


I have also changed the descriptive error snackbars to disappear after 8 seconds instead of being shown indefinitely.
Let me know what you think.

Fs00:
@maniac103 Not sure if you've missed it, but this has been ready for re-review for a while now 🙂 (see https://github.com/slapperwan/gh4a/pull/1433#discussion_r1975331789)

maniac103:
I indeed missed it, sorry for that :-(

Fs00:
No problem, thanks!
Before releasing, please wait for an upcoming PR I'll send this weekend. Today I've found one more crash in the app that seems trivial to fix.

Issue #1252, opened 2022-11-02
[Bug] Link to repo subdirectory loads forever


KaKi87:
Hello,

Clicking a link that points to a repo's subdirectoty displays the loader but gets stuck there forever :

![](https://media.discordapp.net/attachments/723185786794803252/1037500119894999100/Screenshot_20221102-235154_OctoDroid.png)

Then, I have to go back twice before being able to do anything else.

In the above screenshot, I clicked the "source code" link on the following page : https://plugins.jetbrains.com/plugin/14382-deno

Thanks

PR #1438, opened 2025-01-25
Miscellaneous data loading optimizations

Fs00:
This PR implements several optimizations with the goal of reducing loading times (especially in worst-case scenarios) and reduce data usage when possible:
- for issues and PRs, extract comments from timeline API responses instead of making extra calls to the issue comments API. This change alone does not improve loading times, but reduces the amount of fetched data.
- load branches and tags in parallel when opening the branch selection dialog, instead of sequentially fetching branches first and then tags
- employ a smarter mechanism for finding the number of organization members: instead of fetching the full paginated list of members (which can take a long time for orgs with hundreds of public members), we now only fetch the first page by requesting a single item per page, and then look at the number of total pages in the `Link` header.
Since we requested the API to return 1 member per page, the number of pages will be equal to the number of members (except when there are 0 or 1 members in an org, but I made sure to handle that case as well).
- for paginated lists that are fetched entirely by requesting all pages (I looked for usages of `ApiHelpers.PageIterator`), set the number of results per page to 100 instead of using the default (which is 30).
This gives a fairly big performance improvement when paginated lists have several hundreds of items. Below are some examples of such worst-case scenarios:

|  | master | This PR |
|--------|--------|--------|
| branches/tags loading (`torvalds/linux` repo) | 7 secs | 2,5 secs |
| PR conversation loading | 14 secs | 6,5 secs |
| PR files loading (~2000 items) | 22 secs | 8 secs |
| issue conversation loading | 16,5 secs | 8 secs |
| PR review loading | 22,5 secs | 17 secs | 

It's important to note that performance gains are much less prominent for paginated lists that have less items in total (<100). For example, a PR conversation that previously took 6,5 secs to load now takes ~5 secs in my testing.

One last thing worth mentioning: updating OkHttp to version 4 was required to make AS network inspector work properly, since it doesn't support OkHttp 3.x anymore. [Breaking changes in 4.x](https://square.github.io/okhttp/changelogs/upgrading_to_okhttp_4) do not affect us and the bindings were updated to support OkHttp 4 [quite some time ago](https://github.com/maniac103/GitHubSdk/pull/7) :slightly_smiling_face:

### DIFF
--- a/app/src/main/java/com/gh4a/activities/CommitActivity.java
+++ b/app/src/main/java/com/gh4a/activities/CommitActivity.java
@@ -225,7 +225,7 @@ public class CommitActivity extends BaseFragmentPagerActivity implements
 
     private void loadComments(boolean force) {
         final RepositoryCommentService service =
-                ServiceFactory.get(RepositoryCommentService.class, force);
+                ServiceFactory.get(RepositoryCommentService.class, force, ApiHelpers.MAX_PAGE_SIZE);
         ApiHelpers.PageIterator
                 .toSingle(page -> service.getCommitComments(mRepoOwner, mRepoName, mObjectSha, page))
                 .compose(makeLoaderSingle(ID_LOADER_COMMENTS, force))
--- a/app/src/main/java/com/gh4a/activities/CommitActivity.java
+++ b/app/src/main/java/com/gh4a/activities/CommitActivity.java
@@ -224,8 +224,7 @@ public class CommitActivity extends BaseFragmentPagerActivity implements
     }
 
     private void loadComments(boolean force) {
-        final RepositoryCommentService service =
-                ServiceFactory.get(RepositoryCommentService.class, force, ApiHelpers.MAX_PAGE_SIZE);
+        var service = ServiceFactory.getForFullPagedLists(RepositoryCommentService.class, force);
         ApiHelpers.PageIterator
                 .toSingle(page -> service.getCommitComments(mRepoOwner, mRepoName, mObjectSha, page))
                 .compose(makeLoaderSingle(ID_LOADER_COMMENTS, force))
--- a/app/src/main/java/com/gh4a/fragment/UserFragment.java
+++ b/app/src/main/java/com/gh4a/fragment/UserFragment.java
@@ -351,7 +351,7 @@ public class UserFragment extends LoadingFragmentBase implements
     }
 
     private void loadTopRepositories(boolean force) {
-        RepositoryService service = ServiceFactory.get(RepositoryService.class, force, null, null, 5);
+        RepositoryService service = ServiceFactory.get(RepositoryService.class, force, 5);
         final Single<Response<Page<Repository>>> observable;
 
         Map<String, String> filterData = new HashMap<>();
--- a/app/src/main/java/com/gh4a/fragment/UserFragment.java
+++ b/app/src/main/java/com/gh4a/fragment/UserFragment.java
@@ -382,7 +382,7 @@ public class UserFragment extends LoadingFragmentBase implements
             return;
         }
 
-        final OrganizationService service = ServiceFactory.get(OrganizationService.class, force);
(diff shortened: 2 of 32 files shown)


app/src/main/java/com/gh4a/activities/CommitActivity.java
```diff
@@ -225,7 +225,7 @@ private void loadCommit(boolean force) {
 
     private void loadComments(boolean force) {
         final RepositoryCommentService service =
-                ServiceFactory.get(RepositoryCommentService.class, force);
+                ServiceFactory.get(RepositoryCommentService.class, force, ApiHelpers.MAX_PAGE_SIZE);
```
maniac103, line 228: Is there any advantage of passing `MAX_PAGE_SIZE` here over using it as default value in the `get()` call not accepting a page size? Looking at the amount of calls you made use `MAX_PAGE_SIZE` it seems like it should be the new default.

app/src/main/java/com/gh4a/fragment/UserFragment.java
```diff
@@ -396,11 +396,13 @@ private void loadOrganizationMemberCountIfOrg(boolean force) {
...
-                ServiceFactory.get(OrganizationMemberService.class, force);
-        ApiHelpers.PageIterator
-                .toSingle(page -> service.getMembers(mUser.login(), page))
-                .map(memberList -> memberList.size())
+        final OrganizationMemberService service = ServiceFactory.get(OrganizationMemberService.class, force, 1);
+        service.getMembers(mUser.login(), 1)
+                .map(ApiHelpers::throwOnFailure)
+                .map(page -> {
+                    int pagesCount = ApiHelpers.getTotalPagesCount(page);
+                    return pagesCount == 1 ? page.items().size() : pagesCount;
```
maniac103, line 404: Is there any reason this is not just `return pagesCount;`? Since the request is 1 item per page, in what case does the current logic make a difference?

maniac103 (commented):
Impressive improvements :+1:  Got some remarks, but the general approach LGTM.

app/src/main/java/com/gh4a/activities/CommitActivity.java
Fs00: I know it's repetitive, but I'm very reluctant to make it the default for `ServiceFactory.get` because it would also affect all paged lists in the app, which are perfectly fine with GitHub's default of 30 - using `MAX_PAGE_SIZE` for those would only make the initial loading slower.
I agree that it should be the default for `ApiHelpers.PageIterator` methods though. I can try to move the Service creation inside those methods, so that they can set the `pageSize` on their own.

app/src/main/java/com/gh4a/fragment/UserFragment.java
Fs00: When there are 0 members, `pagesCount` is still 1 but there are no items in the list

app/src/main/java/com/gh4a/activities/CommitActivity.java
maniac103: That sounds probably a bit too complicating. I think a good middle ground could be renaming your new method to `ServiceFactory.getForPageIterator(Class<S> serviceClass, boolean bypassCache)` (or something among those lines) and inserting the page size there? That would make the intent clear and still centralize the number at one point. WDYT?

Fs00: I think it's fine. I've named the `ServiceFactory` method `getForFullPagedLists` because I didn't want to couple it too much with the concept of `PageIterator`, but let me know if you have other naming suggestions.

Fs00:
I have also added a commit to leverage `PageIterator.first` in the `RefPathDisambiguationTask` to avoid loading all branches/tags unnecessarily when resolving a blob/tree URL, which should help in situations like the one in #1252.

maniac103:
Thanks 👍

PR #1440, opened 2025-02-02
Minor miscellaneous UI fixes

Fs00:
This PR fixes a few minor UI issues I've spotted throughout the app:

- there were some places where the app didn't parse emoji aliases in text (like `:smile:` or `:+1:`): user bios, label names and repository descriptions in the user/organization screen.
The only place which still lacks emoji parsing is the "Filter by label" selection screen. I've tried to implement it there too, but it broke filtering for labels with emojis because GH search filters expect the "raw" label name (with unparsed aliases). Since fixing this scenario would have required a lot more work, I've left it as is for now.
Also, while fixing repo descriptions in the user screen, I've noticed that in other repository lists there was some logic to display only the first sentence or line of the repo description. In the end, I've decided to get rid of it to make the behavior more predictable and consistent throughout the app.
- embedded videos in GitHub blog posts were overflowing due to some weird inline CSS styles in the article content. I've added a couple of rules to our Markdown stylesheets to override them.
- the diff holder used in the PR review screen had no logic for handling [comments on files](https://github.blog/changelog/2023-04-11-commenting-on-files-in-a-pull-request-is-now-generally-available/) (which have no diff attached to them), and displayed an empty space in place of the diff for those comments:

| Before | After |
|--------|--------|
| Screenshot_20250201-192034_OctoDroid | Screenshot_20250201-192126_OctoDroid_Debug | 

~~Lastly, while working on the diff holder, I've decided to give another try at making the code use the monospace font without wasting too much screen width (I tried to do that [some time ago](https://github.com/slapperwan/gh4a/pull/1144) but wasn't satisfied with the result). I think I have now reached a good compromise, thanks to the use of the `textScaleX` attribute that makes the text a bit more narrow:~~

| Before | After |
|--------|--------|
| Screenshot_20250201-191801_OctoDroid | Screenshot_20250201-191730_OctoDroid_Debug |

### DIFF
--- a/app/src/main/assets/markdown-light.css
+++ b/app/src/main/assets/markdown-light.css
@@ -194,8 +194,12 @@ table tr td :last-child,table tr th :last-child {
     margin-bottom: 0;
 }
 
-img {
+img, video {
     height: auto;
+    max-width: 100%;
+}
+
+.wp-video {  /* Prevents videos from overflowing in GitHub blog articles */
     max-width: 100%;
 }
 
--- a/app/src/main/java/com/gh4a/activities/IssueEditActivity.java
+++ b/app/src/main/java/com/gh4a/activities/IssueEditActivity.java
@@ -75,6 +75,7 @@ import com.meisolsson.githubsdk.service.issues.IssueMilestoneService;
 import com.meisolsson.githubsdk.service.issues.IssueService;
 import com.meisolsson.githubsdk.service.repositories.RepositoryCollaboratorService;
 import com.meisolsson.githubsdk.service.repositories.RepositoryContentService;
+import com.vdurmont.emoji.EmojiParser;
 
 import java.net.HttpURLConnection;
 import java.util.ArrayList;
@@ -991,7 +992,7 @@ public class IssueEditActivity extends BasePagerActivity implements
                 viewColor.setBackgroundColor(ApiHelpers.colorForLabel(label));
 
                 final TextView tvLabel = rowView.findViewById(R.id.tv_title);
-                tvLabel.setText(label.name());
+                tvLabel.setText(EmojiParser.parseToUnicode(label.name()));
                 tvLabel.setOnClickListener(clickListener);
                 tvLabel.setTag(label);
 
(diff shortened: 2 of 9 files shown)


maniac103:
> Lastly, while working on the diff holder, I've decided to give another try at making the code use the monospace font without wasting too much screen width (I tried to do that https://github.com/slapperwan/gh4a/pull/1144 but wasn't satisfied with the result).

Your screenshot still has the same issue you mentioned in #1144 though:

> I didn't use monospace for all the code in the diff hunk for compactness reasons: when code is moderately indented (20+ spaces), you would always need to scroll right to be able to read the text.

Did you experiment with smaller `textScaleX` values (0.9 or 0.8)?

Fs00:
Yeah, I'm aware that the issue is still present, but lowering the `textScaleX` makes it less worse. The monospace font will inevitably be wider than the standard one, but on the other hand the alignment of the diff text is much better.

I've tried with 0.9 and on my device it looks exactly the same as 0.95, tried with 0.85 and the result is the following:


which is probably still acceptable, but below that threshold the text starts looking very stretched (the following is 0.8):

maniac103:
I have yet to try on-device, but to me the 0.8 case still looks acceptable and provides about the same amount of displayed text as the current solution.

Fs00:
It's readable but frankly it looks quite bad to me... we also need to consider that some folks may be looking at the diff on a foldable screen, tablet or in landscape mode, there's no need to stretch the text that much in those scenarios.

> takes the approach a bit further and vastly improves readability of inline diffs in comments :-)

That's what I wanted to do for all `CodeBlockSpan`s in a future PR 😉

maniac103:
> we also need to consider that some folks may be looking at the diff on a foldable screen, tablet or in landscape mode, there's no need to stretch the text that much in those scenarios.

The scaling factor could be put into a float resource so it's only used in the phone-in-portrait scenario...or maybe make it a user preference?

Fs00:
I prefer having it as a float resource (maybe tied to screen width instead of orientation) rather than a preference... we already have the "Code view text size" option which affects the code viewer too, having a similar one with slightly different behavior could be confusing imho.

Anyway, since we'd like to apply a similar (if not the same) appearance to code block spans too, I'd rather iterate on all code-related "look and feel" changes in another PR so that we can unblock the UI fixes here. Is it OK for you?

maniac103:
> I'd rather iterate on all code-related "look and feel" changes in another PR so that we can unblock the UI fixes here. Is it OK for you?

Yes, that sounds like a good idea :+1:

Fs00:
I have removed the monospace font changes and added instead a commit to escape more HTML entities that can be found in GitHub blog post previews.

Issue #1412, opened 2024-11-01
[Proposal] Request app inclusion in the IzzyOnDroid F-Droid repo


Fs00:
I've recently learned about the [IzzyOnDroid repository](https://apt.izzysoft.de/fdroid/index/info), which is an alternative F-Droid repo that distributes original APKs of FOSS apps pulled in directly from their respective source repositories (in our case that would be GitHub releases).
I think that it would be a great alternative for making OctoDroid available to end users, because:
- the app can then be downloaded and updated using existing F-Droid clients, but APKs are the same ones that can be obtained on GitHub
- update availability is faster compared to the main F-Droid repo: as far as I've seen, updates are available in the repository in less than a day once the release is published on GH (F-Droid instead takes *at least* 2 days)
- the APK only needs to be built and uploaded to GH releases and nowhere else

To request the inclusion of the app in the repository, we only need to open an issue [on their Gitlab](https://gitlab.com/IzzyOnDroid/repo). I'm happy to do that if you're fine with it @maniac103.

As far as I've seen, there is only a potential hurdle for inclusion in the repo: they seem to be a bit strict against the `REQUEST_INSTALL_PACKAGES` permission, mainly because it could potentially be used by auto-updaters to *"download additional executable binary files without explicit user consent"*, which would be against the [inclusion policy](https://gitlab.com/IzzyOnDroid/repo/-/wikis/Inclusion%20Policy). Since OctoDroid doesn't fall into that scenario, I'm confident that we can fulfill the requirements for inclusion without the need to remove the permission.
Another thing they look at - for which the repository website displays a warning, but doesn't prevent inclusion - is whether the APK contains the `DEPENDENCY_INFO_BLOCK`, an encrypted/obfuscated signing block added by default by Android build tools ([more info here](https://developer.android.com/build/dependencies#dependency-info-play)) which can be easily removed (I plan to open a PR for that).

Let me know your thoughts!

maniac103:
I'm totally fine with that.
NB: F-Droid can use pre signed APKs as well nowadays (I've [used this](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/15369) recently), but obviously that's too late for Octodroid.

Fs00:
[...]

Issue #1243, opened 2022-10-09
Paged lists behave weirdly after rotating the screen


Fs00:
As reported [here](https://github.com/slapperwan/gh4a/commit/a3edd7cb44a484614976b6d514ecc6907e812d3b#commitcomment-79322386), the fix for #1194 introduced a regression in paged lists:

> I've noticed that if I rotate the screen and then go to the bottom of the paged list, the list suddenly scrolls back to the top.

Issue #948, opened 2020-01-08
Option to Export Bookmarks


natur3kid:
It would be nice to export the Bookmarks and keep them safe somewhere external.

Micha-he:
My approval...

billie80:
It would be great if there is an option to export and import bookmarks. Recreating many bookmarks manually after re-installing or switching devices is quite cumbersome.

PR #1443, opened 2025-02-10
Allow following file renames in commit history activity

### LINKED ISSUE #1419: Better file renaming commit history
Github shows in the file renaming if a file has been changed its name. Octodroid only shows the commit that changed the filename. You can't properly view the commit history of the old file in octodroid. A simple button that allows to view the old commit history should be added.

maniac103:
Fixes #1419

### DIFF
--- a/app/src/main/java/com/gh4a/activities/CommitHistoryActivity.java
+++ b/app/src/main/java/com/gh4a/activities/CommitHistoryActivity.java
@@ -3,22 +3,26 @@ package com.gh4a.activities;
 import android.content.Context;
 import android.content.Intent;
 import android.os.Bundle;
+
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
 
 import com.gh4a.R;
 import com.gh4a.fragment.CommitListFragment;
 import com.meisolsson.githubsdk.model.Commit;
+import com.meisolsson.githubsdk.model.ContentType;
 
 public class CommitHistoryActivity extends FragmentContainerActivity implements
         CommitListFragment.ContextSelectionCallback {
     public static Intent makeIntent(Context context, String repoOwner, String repoName,
-                                    String ref, String path, boolean supportBaseSelection) {
+                                    String ref, String path, ContentType type,
+                                    boolean supportBaseSelection) {
         return new Intent(context, CommitHistoryActivity.class)
                 .putExtra("owner", repoOwner)
                 .putExtra("repo", repoName)
                 .putExtra("ref", ref)
                 .putExtra("path", path)
+                .putExtra("type", type)
                 .putExtra("base_selectable", supportBaseSelection);
     }
 
@@ -26,6 +30,7 @@ public class CommitHistoryActivity extends FragmentContainerActivity implements
     private String mRepoName;
     private String mRef;
     private String mFilePath;
+    private ContentType mType;
     private boolean mSupportBaseSelection;
 
     @Nullable
@@ -47,12 +52,13 @@ public class CommitHistoryActivity extends FragmentContainerActivity implements
         mRepoName = extras.getString("repo");
         mRef = extras.getString("ref");
         mFilePath = extras.getString("path");
+        mType = (ContentType) extras.getSerializable("type");
         mSupportBaseSelection = extras.getBoolean("base_selectable");
     }
 
     @Override
     protected Fragment onCreateFragment() {
-        return CommitListFragment.newInstance(mRepoOwner, mRepoName, mRef, mFilePath);
+        return CommitListFragment.newInstance(mRepoOwner, mRepoName, mRef, mFilePath, mType);
     }
 
     @Override
--- a/app/src/main/java/com/gh4a/fragment/CommitListFragment.java
+++ b/app/src/main/java/com/gh4a/fragment/CommitListFragment.java
@@ -21,10 +21,16 @@ import android.os.Bundle;
 
 import androidx.activity.result.ActivityResultLauncher;
 import androidx.activity.result.contract.ActivityResultContracts;
+import androidx.annotation.NonNull;
+import androidx.core.view.MenuProvider;
 import androidx.recyclerview.widget.RecyclerView;
+
+import android.os.Parcel;
+import android.os.Parcelable;
 import android.view.ContextMenu;
 import android.view.LayoutInflater;
 import android.view.Menu;
+import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.view.View;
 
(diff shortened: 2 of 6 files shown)


maniac103:
@Fs00 Please review

app/src/main/java/com/gh4a/activities/CommitHistoryActivity.java
```diff
@@ -52,14 +57,47 @@ protected void onInitExtras(Bundle extras) {
...
+            followItem.setChecked(mFollowRenames);
+        }
+        return super.onPrepareOptionsMenu(menu);
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        if (item.getItemId() == R.id.follow_renames) {
+            mFollowRenames = !mFollowRenames;
+            item.setChecked(mFollowRenames);
```
maniac103, line 90: Question: store the setting persistently (similarly to 'wrap lines' in file viewer) or keep default as 'off'?

Fs00:
This is an interesting feature, although I have a few doubts:
- why would we need an option to enable/disable it? I think it's very easy to miss the existence of the feature if it's disabled by default, and I don't see why a user would want to deliberately ignore older history entries
- I'm not really knowledgeable about RxJava, but what happens if the user rotates the screen while entries belonging to an "older filename" are being displayed? As far as I can see, the RenameFollowData will be lost upon fragment recreation.

Have you considered implementing this functionality with an action at the end of the list that opens the history of the previous filename (if there is one), similarly to what the GitHub web UI does? I think it could spare us dealing with paged lists and [their known issues](https://github.com/slapperwan/gh4a/issues/1243).

gh_web_file_rename

maniac103:
> why would we need an option to enable/disable it? I think it's very easy to miss the existence of the feature if it's disabled by default, and I don't see why a user would want to deliberately ignore older history entries

It's not the default behavior for `git log` either, but needs to be explicitly enabled (`git log --follow`) there as well.

> I'm not really knowledgeable about RxJava, but what happens if the user rotates the screen while entries belonging to an "older filename" are being displayed? As far as I can see, the RenameFollowData will be lost upon fragment recreation.

That's probably right, need to check this. Could likely put it into the saved state.

> Have you considered implementing this functionality with an action at the end of the list that opens the history of the previous filename (if there is one), similarly to what the GitHub web UI does?

No, but I consider the current solution more suited because
- it's in line with `git log`
- the GH UI solution will get 'interesting' with both the fragment and adapter setup (we'd need to insert a dummy item into the list, but how to do that with the paging setup?)

maniac103:
> That's probably right, need to check this. Could likely put it into the saved state.

Tested this now, added saving of members to saved state, should be fine now.

Fs00:
Great, I'll play with it and take a closer look at the implementation in the next few days 👀

Fs00 (commented):
I've done some tests and in the end I think that the functionality is fine as-is (opt-in without persisting the option in a preference). If it were enabled by default, it would be harder for the user to realize that the file was renamed at some point in the history.
Having a visual indicator on those commits in which the file is renamed (like a _"Renamed to X"_ chip) would address this inconvenience, but I don't feel it's necessary at the moment (could be a future improvement).

While testing I've only found two minor issues (except the known bug #1243):
- after enabling the option and rotating the screen, the "Follow renames" checkbox turns back to being disabled. I guess it's because the state is saved in the fragment but the menu is created in the activity.
Wouldn't it be better to have the options menu being inflated in the `CommitListFragment` only when a `filePath` is provided? That way we wouldn't even need to touch the `CommitHistoryActivity`.
- following renames for directories does not work, but I believe it's a Git limitation because it doesn't work on the GitHub web UI either. Would it be possible to hide the "Follow renames" option on folders?

app/src/main/java/com/gh4a/activities/CommitHistoryActivity.java
Fs00: I think it's fine as-is, see my comment below

app/src/main/java/com/gh4a/fragment/CommitListFragment.java
```diff
@@ -164,4 +223,69 @@ protected Single<Response<Page<Commit>>> loadPage(int page, boolean bypassCache)
...
     }
+
+    @Override
+    protected void resetSubject() {
+        super.resetSubject();
+        mRenameFollowData = null;
+        mOldestLoadedCommit = null;
+    }
+
+    private Maybe<RenameFollowData> determineFollowData(
```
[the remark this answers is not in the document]
maniac103: Named it `determineFollowDataIfRenamedInCommit` and `FileRenameFollowData` now. Want to keep the 'follow' aspect, as that's  the point of the class (tracking data required to follow a rename in the git log).

Fs00:
Renames look good to me, let me know when the two issues are addressed

maniac103:
>let me know when the two issues are addressed

What issues do you mean here?

Fs00:
I mean the two bullet points in my review above

maniac103:
> I mean the two bullet points in https://github.com/slapperwan/gh4a/pull/1443#pullrequestreview-2778851991

Oops, sorry, completely missed those. Fixed both of them.

app/src/main/java/com/gh4a/fragment/CommitListFragment.java
```diff
@@ -92,13 +108,55 @@ public void onCreate(Bundle savedInstanceState) {
...
+    public void onSaveInstanceState(@NonNull Bundle outState) {
+        super.onSaveInstanceState(outState);
+        outState.putBoolean("follow_renames", mFollowRenames);
+        outState.putParcelable("rename_follow_data", mMostRecentRenameFollowData);
+        outState.putParcelable("oldest_commit", mOldestLoadedCommit);
+    }
+
+    @Override
+    public void onCreateMenu(@NonNull Menu menu, @NonNull MenuInflater inflater) {
+        if (mFileType == ContentType.File) {
```
Fs00, line 129:
```suggestion
        if (mFileType == ContentType.File || mFileType == ContentType.Submodule) {
```
I've just checked with submodules [in this folder](https://github.com/topjohnwu/Magisk/tree/master/native/src/external) and following renames works fine for them too.
At the moment this change wouldn't be strictly necessary since submodules are returned with their type set as `file`, but in a future REST API version GH might change that according to the [API docs](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content):
> In the next major version of the API, the type will be returned as "submodule".

Fs00:
Great! 🎉
If you are going to tag a release soon, please remember to [change and rotate the APK signing key](https://anandbose.dev/android-signing-key-rotation-explained/) otherwise the app [cannot be included in the IzzyOnDroid repo](https://github.com/slapperwan/gh4a/issues/1412#issuecomment-2517080654).

maniac103:
> please remember to [change and rotate the APK signing key](https://web.archive.org/web/20241218183439/https://anandbose.dev/android-signing-key-rotation-explained/)

I am aware, but I'm procastinating there because I don't fully understand the details of the process yet and am not sure of the implications, e.g. for how long is this double signing needed - only once?
Do you have experience on that process?

Fs00:
No, I don't have any experience.

> how long is this double signing needed - only once?

That's a good question - my understanding is that it's needed as long as you want to be able to upgrade from an APK signed with an old key, but I couldn't find it explicitly stated anywhere.
However, after a bit of research I've found out that things have become more complicated since Google introduced the [APK signature scheme v3.1](https://source.android.com/docs/security/features/apksigning/v3-1) in Android 13 to address rotation handling issues in the OS. In short, the default apksigner behavior now involves signing an APK with both the old key (used for Android <= 12L) and the rotated one (for Android 13+), which is quite messy to be honest.

I'm not too sure if rotating the key is worth at this point, considering that many people have likely downloaded the app from F-Droid. The only potentially frustrating thing when reinstalling the app from scratch would be to re-add previous bookmarks...

xabolcs:
> The only potentially frustrating thing when reinstalling the app from scratch would be to re-add previous bookmarks...

How about implementing a nice "export/import OctoDroid settings" feature?

Fs00:
I see that there is already #948 for requesting the import/export bookmarks functionality, I think that one is enough. Imho I don't see much value in a settings backup feature considering the limited number of options the app has.

Btw @maniac103 if the signing key rotation ends up being too complex to handle, feel free to release without changing the key.

PR #1444, opened 2025-02-10
Remove Optional class in favor of java.util.Optional

maniac103:
When using Java streams dealing with two different Optional classes becomes confusing otherwise.


maniac103:
@Fs00 Thoughts on this? Unfortunately there still is a naming conflict between `java.util.functions.Function` (doesn't throw) and `io.reactivex.functions.Function` (can throw).

Fs00:
Last time I didn't even realize that I wasn't using the standard library `Optional<T>` 🙈

You might have guessed that I'm always in favor of these cleanup refactorings ;)
The only thing I'd do is renaming `RxUtils.flatMapToSingle` to `mapToSingle` because there's actually no "flattening" involved (the mapper returns a plain `Single<T>` and not a `Single<Optional<T>>`), everything else looks great.

PR #1446, opened 2025-02-15
Implement basic support for Markdown alerts in HTML-to-spanned conversion

Fs00:
This PR adds basic rendering support for [Markdown alerts](https://github.com/orgs/community/discussions/16925), a special type of blockquote that has been around for one year now.

The appearance is kept intentionally simple (no colors, no title icons) to avoid dealing with different color tones for dark/light themes and to make them visually consistent with how they appear in the WebView viewer, where they are displayed as simple blockquotes because the Markdown JS library we use doesn't yet support them.

| Before | After |
|--------|--------|
| alert_no_support | alert_with_support |

### DIFF
--- a/app/src/main/java/com/gh4a/utils/HtmlUtils.java
+++ b/app/src/main/java/com/gh4a/utils/HtmlUtils.java
@@ -59,9 +59,12 @@ import org.xml.sax.XMLReader;
 
 import java.io.IOException;
 import java.io.StringReader;
+import java.util.Arrays;
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


Fs00:
Found a bug when an alert contains a `<div>`, will look into it before marking it as ready.

Fs00:
I've had to rework the approach I had initially taken.
We now keep track whether an element is a Markdown alert or an alert title on the mark used for block elements (which was previously called `Newline`). If we didn't do that, we'd have no way to tell whether an ended `<div>` is the end of a Markdown alert or simply a `<div>` nested inside the alert (a real world example is the alert at the top of [this README](https://github.com/facebook/create-react-app/blob/main/README.md)).

While I was at it, I've also converted most trivial mark classes to records and fixed a spacing issue when rendering consecutive tables.

